PR #138, opened 2014-12-13
more refactoring: StringStreamer

bluej100:
I'm not positive this is correct, but it feels like an improvement to me, and it passes the tests.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -52,6 +52,7 @@
 	global.Papa.ParserHandle = ParserHandle;
 	global.Papa.NetworkStreamer = NetworkStreamer;
 	global.Papa.FileStreamer = FileStreamer;
+	global.Papa.StringStreamer = StringStreamer;
 
 	if (global.jQuery)
 	{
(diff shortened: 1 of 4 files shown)


mholt:
Hm, I'm seeing failing tests at the moment, but maybe it's just a rebase issue; the tests that are failing are ones that were introduced with recent PRs.

bluej100:
Hmm, yup. I've pushed up a rebased patchset with fixes for those tests.

I fixed it by switching another test from complete to setTimeout, though, which is gross. How do you expect the complete callback to behave with pausing and aborting? I believe that currently, it fires immediately after aborting, which seems weird to me.

mholt:
Complete should be executed when parsing is finished, whether aborted or not. It shouldn't be executed after pausing...

The meta object should contain an `aborted` property to indicate that parsing was aborted. But I think it's still important that complete is called when parsing is aborted so the user can get the partly-finished results.

The test "Step functions can resume parsing" seems to be failing; not sure why yet, since I've confirmed that the handle's resume function *is* being invoked and the parser is parsing the remaining input.

bluej100:
Okay, I'll take a look. Thanks!

bluej100:
Sorry for the delay. New patchset pushed which implements the complete callback behavior you specified and adds some more robustness to the custom test runner. Hope it meets your approval.

bluej100:
It's kind of confusing to me that streamers and handlers can be paused separately. Might want to unify that.

mholt:
Whew, finally got through the last big one! Thanks for all your time and effort. A lot of these issues have been thorns in my side.

Hm, what do you mean by streamers and handles being paused separately?

bluej100:
ChunkStreamer has its own resume/pause/abort functions, which used to be in NetworkStreamer and FileStreamer. I'm guessing that no one is using them, since people don't generally construct streamers directly. We should probably either kill them or add testing and documentation.

mholt:
I see what you mean now. I've ripped those out in a commit that's on its way. Thanks again!

PR #139, opened 2014-12-13
use fast mode automatically when safe

bluej100:


bluej100:
Rebased and restored the explicit config.fastMode flag.

bluej100:
Rebased. Last one! Thanks for all your work.

bluej100:
Yay! You're on a rampage tonight.

PR #149, opened 2015-01-05
Making Papa require friendly.

mnibecker:
Added functionality to return Papa obj if app is using require.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -1432,4 +1432,11 @@
 	{
 		return typeof func === 'function';
 	}
+
+	// if requirejs is available, return Papa
+	if (typeof define === "function" && define.amd) {
+	    define('papaparse', [], function() {
+	    	return global.Papa;
+    	});
+	}
 })(this);
--- a/papaparse.js
+++ b/papaparse.js
@@ -1434,9 +1434,7 @@
 	}
 
 	// if requirejs is available, return Papa
-	if (typeof define === "function" && define.amd) {
-	    define('papaparse', [], function() {
-	    	return global.Papa;
-    	});
+	if (isFunction(define) && define.amd) {
+	    define( function() { return global.Papa; });
 	}
 })(this);


mnibecker:
Additionally I was thinking about add an else to return set `global.Papa = Papa` and not have Papa automatically inserted into the global namespace (which I think is more correct), but I'm concerned about this breaking some require apps that just import using `window.Papa`

papaparse.js
```diff
@@ -1432,4 +1432,11 @@
 	{
 		return typeof func === 'function';
 	}
+
+	// if requirejs is available, return Papa
+	if (typeof define === "function" && define.amd) {
```
mholt: For consistency, let's use isFunction (defined just above your edit, actually). And if `define` is defined on the `window` object (accessible as `global` inside the Papa closure), let's explicitly say `global.define` to make it clear.

mholt:
Thanks for the PR -- I've never used RequireJS so I'll trust you to make the integration that's best.

Sorry it took me a while to get around to this.

As for setting `global.Papa = Papa`, I don't think that's necessary; `global` is `window` (at least, in the browser environment).

papaparse.js
```diff
@@ -1432,4 +1432,11 @@
...
+	    define('papaparse', [], function() {
```
fearphage: Is there a reason to name it? Why can't it be an anonymous define?

```` javascript
define(function() {
  return global.Papa;
});
````

File: papaparse.js [hunk 3]
```diff
@@ -1432,4 +1432,11 @@
...
+	    	return global.Papa;
+    	});
```
fearphage: Indentation is off.

papaparse.js
mnibecker: good call, I'll leave it anonymous

mnibecker:
Sorry it took me so long to get this. Made the recommended changes

mholt:
Thanks!

PR #166, opened 2015-02-04
remove references to streamer._paused and fix fake chunk callbacks

### LINKED ISSUE #164: There is an issue with .pause/.resume
I heavily rely on pause-resume methods. I use chunk-streaming processing and here is what I've found:
https://github.com/mholt/PapaParse/blob/master/papaparse.js#L403

```javascript
else if (isFunction(this._config.chunk))
{
	this._config.chunk(results, this._handle);
	if (this._paused)
		return;
	results = undefined;
}
```

`this._paused` is always false in this context. It has to be checked by `this._handle.paused()`

Another part of the issue is in the resume implementation, but let's look at `pause` method at first:

```javascript
this.pause = function()
{
	_paused = true;
	_parser.abort();
	_input = _input.substr(_parser.getCharIndex());
};
```

`_input` will be always an empty string (!) if we use chunk-streaming. On resume, PapaParse will parse the empty string again and will fire callbacks with empty results.

My JavaScript code is not that good, which is why I'm creating this issue instead of a pull request, but here is my attempt to patch the issue:

```diff
372a373,378
>             
>             if (aggregate.length === 0) {
>                 // It is possible due to the pause/resume implementation.
>                 this._nextChunk();
>                 return;
>             }
403c409
<                               if (this._paused)
---
>                               if (this._handle.paused())
```

bluej100:
What do you think of this? It would give weird behavior if someone used step and chunk callbacks simultaneously, but since that isn't supported, I think it's reasonable.

I think the only behavior difference with this patch is that before, if you paused after the final chunk callback, it would still call the complete callback immediately.

Fixes #164 . Pardon the delay and my introducing the bug in the first place.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -355,7 +355,6 @@
 	function ChunkStreamer(config)
 	{
 		this._handle = null;
-		this._paused = false;
 		this._finished = false;
 		this._input = null;
 		this._baseIndex = 0;
@@ -365,7 +364,7 @@
 		this._nextChunk = null;
 		replaceConfig.call(this, config);
 
-		this.parseChunk = function(chunk)
+		this.parseChunk = function(chunk, isFakeChunk)
 		{
 			// Rejoin the line we likely just split in two by chunking the file
 			var aggregate = this._partialLine + chunk;
@@ -397,15 +396,15 @@
 					finished: finishedIncludingPreview
 				});
 			}
-			else if (isFunction(this._config.chunk))
+			else if (isFunction(this._config.chunk) && !isFakeChunk)
 			{
 				this._config.chunk(results, this._handle);
-				if (this._paused)
+				if (this._handle.paused())
 					return;
 				results = undefined;
 			}
 
-			if (finishedIncludingPreview && isFunction(this._config.complete) && (!results || !results.meta.aborted))
+			if (finishedIncludingPreview && isFunction(this._config.complete))
 				this._config.complete(results);
 			
 			if (!finishedIncludingPreview && (!results || !results.meta.paused))
@@ -723,14 +722,15 @@
 		this.resume = function()
 		{
 			_paused = false;
-			self.streamer.parseChunk(_input);
+			self.streamer.parseChunk(_input, true);
 		};
 
 		this.abort = function()
 		{
+			_paused = true;
 			_parser.abort();
 			if (isFunction(_config.complete))
-				_config.complete(_results);
+				_config.complete({ meta: { aborted: true } });
 			_input = "";
 		};
 
--- a/tests/test-cases.js
+++ b/tests/test-cases.js
@@ -1159,6 +1159,72 @@ var CUSTOM_TESTS = [
 		}
 	},
 	{
+		description: "Chunk functions can pause parsing",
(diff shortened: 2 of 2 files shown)


bluej100:
Hmm, I'm going to add an abort test as well. Jussec.

bluej100:
Updated.

frol:
I don't really like either isFakeChunk name or the approach for the fix. IMHO, The code does a lot of meaningless manipulations (strings concatenation, parsing, check for something irrelevant in the context of empty input) where it had to just skip the chunk.

bluej100:
It doesn't just have to skip the chunk. It also needs to call the complete
callback if it was the last chunk, for example.

frol:
How about to check whether it is completed before pausing? Then on resume you won't need to make this empty parsing.

papaparse.js
```diff
@@ -723,14 +722,15 @@
...
+			self.streamer.parseChunk(_input, true);
 		};
 
 		this.abort = function()
 		{
+			_paused = true;
 			_parser.abort();
 			if (isFunction(_config.complete))
-				_config.complete(_results);
+				_config.complete({ meta: { aborted: true } });
```
mholt: I actually think this is fixed with the latest (4.1.1).

papaparse.js
```diff
@@ -397,15 +396,15 @@
...
 			{
 				this._config.chunk(results, this._handle);
-				if (this._paused)
+				if (this._handle.paused())
 					return;
 				results = undefined;
 			}
 
-			if (finishedIncludingPreview && isFunction(this._config.complete) && (!results || !results.meta.aborted))
+			if (finishedIncludingPreview && isFunction(this._config.complete))
```
mholt: Does this duplicate the complete callback? We call it in the abort function too.

papaparse.js
```diff
@@ -397,15 +396,15 @@
 					finished: finishedIncludingPreview
 				});
 			}
-			else if (isFunction(this._config.chunk))
+			else if (isFunction(this._config.chunk) && !isFakeChunk)
 			{
 				this._config.chunk(results, this._handle);
-				if (this._paused)
+				if (this._handle.paused())
```
mholt: Definitely keep this :+1:

papaparse.js
```diff
@@ -355,7 +355,6 @@
 	function ChunkStreamer(config)
 	{
 		this._handle = null;
-		this._paused = false;
```
mholt: This is a good change - I think we have consensus that this variable is not used. :+1:

mholt:
@bluej100 So now that it's 7 months later (my bad!) how do you feel about this PR?

bluej100:
I think I had a fancy idea for having ParserHandler remember where it left off and directly resume instead of sending a fake chunk to the streamer. Perhaps we shouldn't let the perfect be the enemy of the good here--this seems like an improvement to me that probably shouldn't be blocked in hopes of getting me to do something better.

gabegorelick:
I've taken @bluej100's great work and updated it to work on `master`. See #479.

pokoli:
I'm closing this in favour of #479

Issue #167, opened 2015-02-06
Papaparse not working with browserify


ChrisAlvares:
Because of the way browserify module's work, the scope [here](https://github.com/mholt/PapaParse/blob/347befde6c8c6c4f23b0097b32e448e34fe28d15/papaparse.js#L1337 ) is not actually window. 

While the ```this``` keyword is good for keeping everything inclosed, the papaparse library uses the ```window.postMessage``` function, so the anonymous function must return ```window```,  otherwise you run into errors.

Before I submitted a PR, I wanted to get @mholt's opinion on if this is the right move here.

mholt:
So just replacing `this` on line 1337 (see what we did there?) with `window`? I imagine that would be fine. Go ahead and submit a PR and we'll double-check.

*says the guy who's never used Browserify or RequireJS, or other Javascript modules/build tools, etc...*

bluej100:
Sounds good to me. Thanks.

robd:
I've been [getting PapaParse to work with webpack](http://stackoverflow.com/questions/28561396/loading-papaparse-dynamically-with-webpack), and I found I needed to use the [import and export loaders](http://webpack.github.io/docs/shimming-modules.html):

`require("imports?this=>window!exports?global.Papa!./papaparse.js")`

But if I change the script to pass `window` instead of `this` as suggested, I no longer need the import loader part (ie `imports?this=>window`). So I think this would be an improvement for webpack integration too.

mholt:
How would changing `window` to `this` affect Node environments?

robd:
I'm not sure I'm afraid. Guessing, I would say that the window object probably doesn't exist on the server side, so it would probably break. This seems to be being addressed here in BabyParse on [L845](https://github.com/Rich-Harris/BabyParse/blob/b7a7ef850d5720d17689e3df4b3daba5bbc9f965/babyparse.js#L845). 

Also [lines 830-843](https://github.com/Rich-Harris/BabyParse/blob/b7a7ef850d5720d17689e3df4b3daba5bbc9f965/babyparse.js#L830-L843) seem to handle returning the Papa object correctly for different module systems. I tried retrofitting these into PapaParse, and replacing calls to global.Papa with Papa - https://github.com/robd/PapaParse/compare/webpack-integration?expand=1. This was successful; it removed the need for any webpack loader shims and stopped Papa from being added to the window namespace when loaded as a module. 

Seems to make sense, and tests pass, but not sure whether this breaks anything. 

[...]

PR #172, opened 2015-02-18
Module loading

robd:
Further to discussions in #167, this PR adds support for loading PapaParse as a module. It adds basic automated tests for loading as a node js module (although some options will not work eg remote file parsing, web workers etc). It also adds support for loading via RequireJS (untested) and WebPack (manually tested). Finally, it stops PapaParse from being added into the global context (ie window) when loaded as a module.


bluej100:
http://awesomegifs.com/wp-content/uploads/carl-sagan-youre-awesome.gif

PR #178, opened 2015-03-03
fixed support for aborting parsing

JohnMacAslan:
I was using PapaParse to parse through a large CSV file recently, and wanted to make use of the ParseHandle's abort method. However, despite calling abort(), the parser continued to run. Following the pattern for pausing the parser, I attempted to flesh out support for calling abort. From what I could tell, there just wasn't a variable in place for abort() like there is for pause(). 

The original test case seemed to pass regardless of whether or not abort() was working, because the string provided was being parsed in a single go. Hence, the callback was only called once no matter what. I included the chunkSize on the config to make the test parse the string in pieces. Now, the test does indeed fail when abort() is not working, and passes when it does work.

Feedback is welcome.

### DIFF
--- a/tests/test-cases.js
+++ b/tests/test-cases.js
@@ -1234,6 +1234,22 @@ var CUSTOM_TESTS = [
 		}
 	},
 	{
+		description: "Complete is called after aborting",
+		expected: true,
+		run: function(callback) {
+			var updates = [];
+			Papa.parse('A,b,c\nd,E,f\nG,h,i', {
+				step: function(response, handle) {
+					handle.abort();
+				},
+				chunkSize: 6,
+				complete: function (response) {
+					callback(response.meta.aborted);
+				}
+			});
+		}
+	},
+	{
 		description: "Step functions can pause parsing",
 		expected: [['A', 'b', 'c']],
 		run: function(callback) {
--- a/tests/test-cases.js
+++ b/tests/test-cases.js
@@ -1227,10 +1227,9 @@ var CUSTOM_TESTS = [
 				step: function(response, handle) {
 					updates.push(response.data[0]);
 					handle.abort();
-				},
-				complete: function() {
 					callback(updates);
-				}
+				},
+				chunkSize: 6
 			});
 		}
 	},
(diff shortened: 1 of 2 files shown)


mholt:
Sorry it's taking me a while to look at this John, I've been flooded with schoolwork. I'll get around to it soon enough!

tests/test-cases.js
```diff
@@ -1227,10 +1227,9 @@ var CUSTOM_TESTS = [
 				step: function(response, handle) {
 					updates.push(response.data[0]);
 					handle.abort();
-				},
-				complete: function() {
 					callback(updates);
```
bluej100: Matt told me at one point that he thinks the complete callback should still be sent after an abort. Would it be difficult to preserve that behavior?

mholt: It looks like the code still calls complete (just glancing at the diff; haven't had a chance to pull it down) but yeah, the test should still assert that complete is being called, even for aborted tasks. In the case of an aborted parse, the meta property `aborted` should be `true`.

mholt:
Also, we should probably make sure this change still jives well (meaning, the test still passes only in the right conditions) after the latest merge today.

tests/test-cases.js
JohnMacAslan: Complete is supposed to be called, but according to the documentation, "When streaming, parse results are not available in this callback." So I was using the callback inside to step function to verify the initial results, and then if the abort failed, the test case would fail with '(multiple results from test)'. I could make another test for verifying that complete is still called.

mholt: Ah, that makes sense -- that way we can actually check the results for correctness, not only that complete was called. If you could add that one other test, that would be stupendous!

@bluej100 Feel free to take a look at that when it comes in; also you may merge when you think it is ready and correct... thank you!

bluej100: Yeah, a separate test sounds like a good idea.

JohnMacAslan: I've added a test to verify that complete is called after aborting. Unless you see anything else, this should be ready to go.

bluej100:
I've pulled and rebased this locally. Looks good to me. Thanks, @JohnMacAslan .

Issue #179, opened 2015-03-08
Complete callback it's receiving partial results on long csv


topicus:
I've tried to parse a long csv (1.8 millon of rows) and only the last chunk of parsed data it's received by the complete callback. 

### Steps to reproduce it.
- [ ] Go to http://papaparse.com/demo
- [ ] Download large csv file
- [ ] Use local file read method
- [ ] Press parse
- [ ] Row count is 179332 instead of 1094353 which is the real number of line rows.

topicus:
I found the problematic commit using git bisect. 

https://github.com/mholt/PapaParse/commit/0785501b30b61aea9ab0a96dd2b8b8cdb9768260

I will try to fix it.

mholt:
Oops! I guess we need tests for large files. I will try to find a good way to test those.

Thank you for looking into it on your own. :clap: I know I'm too busy until the semester ends to get around to it. Let me know if you have any questions along the way; I'll try to be of help.

topicus:
I found the error but I'm not completely sure about the solution. https://github.com/mholt/PapaParse/blob/master/papaparse.js#L400

Just changing that line from:

```javascript
this._partialLine = aggregate.substring(lastIndex - this._baseIndex);
```
To this
```javascript
this._partialLine = aggregate.substring(this._baseIndex - lastIndex);
```
does the trick but I'm concerned about how this should work. In this way we are parsing over and over the same information. 

It would be better to have a sort of accumulator to store results.

I think FileStreamer was modified to parse by chunks but parseChunk wasn't prepared to allow that. 

I figured out this because parse method of ParseHandler constructor on the 4.0.7 version is executed only one time when I select a big file. By contrast, in the 4.1 version parse method is triggered as many times as chunks there are. 

Maybe you could make a brief explanation how this should work.

mholt:
@bluej100 Would you have a chance to answer his questions? I would but school is keeping me way too busy right now.

bluej100:
Yes, I will look into this tomorrow. Sorry.

bluej100:
I've created a pull request. @mholt , I can leave it for your review (30 lines) or merge it as you prefer.

bluej100:
By the way, @topicus , I appreciate that you dove into the code and suggested a fix. I went a different direction for now, restoring the previous behavior of scanning the entire file in one chunk, but I do like the accumulator as an idea for the future.

mholt:
[...]

PR #181, opened 2015-03-13
read files in a single chunk for complete results (#179)

bluej100:
This restores the unchunked read previously present in https://github.com/mholt/PapaParse/commit/57a7349c41502afff0328bab918d0e10f8b8fd80#commitcomment-10179215 .


mholt:
Yes! Thank you!

bluej100:
Thanks for your patience with my spotty participation. I really like this project.

mholt:
Well, I'm kind of out of commission until this semester ends. Since I'll have to get burnt out writing code either way, I might as well get burnt out with school instead of open source. So for a while, it's all up to the you and the contributors. Great system, actually!

PR #183, opened 2015-03-16
Issue#8 Fix complete results with partial results

topicus:
This PR was created based on this commit b3c84bf1f1a0781e5d30fe91e44c9e23699c362b. Take a look at this and let me know if you see something wrong.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -380,6 +380,11 @@
 		this._rowCount = 0;
 		this._start = 0;
 		this._nextChunk = null;
+		this._completeResults = {
+			data: [],
+			errors: [],
+			meta: {}
+		};
 		replaceConfig.call(this, config);
 
 		this.parseChunk = function(chunk)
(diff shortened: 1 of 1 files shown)


bluej100:
This looks sensible to me. We'll also need to remove the logic in FileStreamer and ChunkStreamer that skips chunking if neither chunk nor step is defined.

topicus:
Maybe I missing something but I could found that logic. Code in this PR works accumulating the results only if user does not define step nor chunk.

If user define step or chunk then results are not stored in this._completeResults. 

This is useful when you have a really big CSV and you don't want you memory blows up. 

On the other hand when you have a medium size CSV (< 500MB) handle all the file in memory make sense.

Currently FileStreamer is chunking by default and because max input length string is implementation dependant and we couldn't know how big is this value.

Anyway, probably you could give me a clue because I do not fully understand all the code yet.

bluej100:
I'm referring to these lines on master, which deactivate streaming:

			if (!config.step && !config.chunk)
				configCopy.chunkSize = null;

Would you mind rebasing and removing those two lines? I appreciate your time.

topicus:
Done. I've also ran the test and all passed. Let me know if you need more help.

bluej100:
Looks great. Thank you.

Issue #186, opened 2015-03-19
Skip lines at the beginning of a file? (Feature Suggestion)


zemzelett:
Hi there,

just deleted two big paragraphes of rambling why i want that feature, instead i'll give the same in short:

I wondered if you ever thought of a parse config object option called _startAtLine: number_ which then tells papaparse at which line the actual csv data starts if there are lines in a file (at the beginning), that are not csv data (e.g. general information about the data).

Is this a feature you might add in the future or could you think of doing so?

mholt:
This could be useful for ARFF files (popular in some areas of machine learning especially the academic realm). I will take a closer look at your PR when I have time, and that might answer my question, but: Why not just truncate the beginning of the input string before you parse? Is it really Papa's job to skip lines, or should the input be prepared for the parser before being submitted to it? Or maybe the extra lines at the beginning could be commented out? Papa can skip lines starting with a certain string.

bluej100:
> Is it really Papa's job to skip lines, or should the input be prepared for the parser before being submitted to it?

Currently, if users want to manipulate the input, then they can't take advantage of Papa's file / URL chunk handling, which is super useful. As I suggested in the linked thread, it might be nice to expose a callback for users to prepare chunks.

mholt:
Ah, good point. As you can tell I was a little :sleeping: when I wrote that.

So a callback to prepare chunks - that's interesting. What would that look like? Receives the string to be parsed and returns the *actual* string to parse, maybe? That could be pretty simple. I'd like to see what that looks like.

bluej100:
Yeah, precisely. I'm not sure whether it should happen before or after the trailing line from the previous chunk is prepended. Maybe it should only be called for the first chunk, even.

mholt:
Since chunks are somewhat arbitrarily sized, I'm not sure that there would be much value in having a preprocess callback for each chunk - with workers, especially, it'll just involve lots of string copying. So you have a good point, we may only need this callback to execute once, at the beginning.

bluej100:
Yeah, that seems reasonable. It fails if you have more than a chunk of preface, but that would be kind of ridiculous.

mholt:
And since the chunk size is customizable, it would be the user's responsibility to be aware of that. Sounds good to me so far!

zemzelett:
I never heard of the term ARFF before but  i assume your where talking about this all the time? (Considering ARFF files hold meta information about the data being presented)

[...]

PR #187, opened 2015-03-19
Adding feature startAtLine code (Feature Request from Issue #186)

zemzelett:
Quick addition of the feature. I did test it with a few files quickly.

Maybe this gives a hint to what i meant on the Issue i opened.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -946,6 +946,7 @@
 		var step = config.step;
 		var preview = config.preview;
 		var fastMode = config.fastMode;
+        var startAtLine = config.startAtLine ? config.startAtLine - 1 : 0; //Due to 0-base
 
 		// Delimiter must be valid
 		if (typeof delim !== 'string'
@@ -993,7 +994,7 @@
 			if (fastMode || (fastMode !== false && input.indexOf('"') === -1))
 			{
 				var rows = input.split(newline);
-				for (var i = 0; i < rows.length; i++)
+				for (var i = 0 + startAtLine; i < rows.length; i++)
 				{
 					var row = rows[i];
 					cursor += row.length;
@@ -1022,7 +1023,13 @@
 				return returnable();
 			}
 
-			var nextDelim = input.indexOf(delim, cursor);
+            if(startAtLine) {
+                for(var i = 0; i < startAtLine; i++) {
+                    cursor = input.indexOf(newline, cursor) + 1;
+                }
+            }
+
+            var nextDelim = input.indexOf(delim, cursor);
 			var nextNewline = input.indexOf(newline, cursor);
 
 			// Parser loop
--- a/papaparse.js
+++ b/papaparse.js
@@ -1024,8 +1024,28 @@
 			}
 
             if(startAtLine) {
+                var dQuoteDelim = '"' + delim;
+                var dQuoteNewline = '"' + newline;
+
                 for(var i = 0; i < startAtLine; i++) {
-                    cursor = input.indexOf(newline, cursor) + 1;
+                    var nextDQuotes = input.indexOf('"', cursor);
+                    var nextDQuoteDelim = input.indexOf(dQuoteDelim, cursor);
+                    var nextDQuoteNewline = input.indexOf(dQuoteNewline, cursor);
+                    nextNewline = input.indexOf(newline, cursor);
+
+                    if(nextDQuotes === -1 || nextDQuotes > nextNewline) {
+                        cursor = nextNewline + newline.length;
+                        continue;
+                    }
+
+                    if(nextDQuoteDelim !== -1 && nextDQuoteNewline > nextDQuoteDelim || nextDQuoteNewline === -1) {
+                        cursor = nextDQuoteDelim + dQuoteDelim.length;
+                        i--;
+                        continue;
+                    }
+
+                    //newline should be right after the quotes now resulting in a new record
+                    cursor = nextDQuoteNewline + dQuoteNewline.length;
                 }
             }
 
--- a/tests/test-cases.js
+++ b/tests/test-cases.js
@@ -425,6 +425,51 @@ var CORE_PARSER_TESTS = [
 			data: [['a', 'b', 'c'], ['']],
 			errors: []
 		}
+	},
+    {
+		description: "startAtLine feature with fastMode",
+		input: 'a,b,c\nd,e,f',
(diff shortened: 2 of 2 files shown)


mholt:
Thanks for the PR! Me and @bluej100 will take a look at this. Since it has changes to the core parser I will also need to run some benchmarks and verify that there isn't a slowdown (although, I doubt it just from a glance).

bluej100:
@zemzelett , would you mind adding a test for this in tests/test-cases.js to ease maintenance?

zemzelett:
I'll try to find some time to look into writing a test for this.

papaparse.js
```diff
@@ -1022,7 +1023,13 @@
 				return returnable();
 			}
 
-			var nextDelim = input.indexOf(delim, cursor);
+            if(startAtLine) {
+                for(var i = 0; i < startAtLine; i++) {
+                    cursor = input.indexOf(newline, cursor) + 1;
```
bluej100: This actually will not work, since there could be new lines in quotes.

bluej100:
I would be happy to implement this myself if you don't have time. Just don't want to steal your credit.

Arguably users can handle this problem themselves, but I think it makes sense as a complement to the preview option, which users could also handle themselves by aborting.

zemzelett:
Just updated the code with some test cases.

The code ignores incorrect data on the lines that are getting skipped. It just checks for _newlines_ in quotes.

The code should definitely be reviewed. I don't know if my logic is 100% correct. Will review it myself tomorrow.

bluej100:
I appreciate your thorough work here. It's clever that you were able to do this without changing either of the main parser loops at all.

You could consider doing this at a higher level by running an initial parse with the preview option set to the value of the startAtLine option. Then you could use the cursor position that comes back from that parse as the starting position of the real parse. It might make for simpler code. What do you think?

zemzelett:
Well. I would never have thought of this myself as don't know the code all to well. I'm gonna look at what you suggested.

Btw: i avoided touching the main Parser loops for two reasons: simplicity and readability of the code and i think the main parser loops would be slowed down by the additional code.

zemzelett:
1. In short
----
Ok, i just wrote a lot of text to explain things now realizing, that you might did not read, what my feature request was about (#186).

Actually i did just want to skip lines at the beginning of a file __not assuming it to be csv data__, which my first pull request just did without problems.
But i also realize it might not be so easy to add this feature without deciding what it is for. 

Read the second part for an answer to your last question / request.
Read the third part for a little analysis of what the feature could be about.

2. tl/dr warning!
---

It might be, that i could not grasp your thoughts but when i utilize the preview feature to implement skipping the first _n_ lines, i imagined encountering the following disadvantages:
* a step function would be called on the data, which is supposed to be skipped
* we'd be fumbling around in the main parser; possibly slowing it down when it's not necessary

On a higher level means in <code>this.parse</code> (line 705) of the <code>ParseHandle</code>? Might be good, avoiding extra code in the main parser. Still the disadvantage of the step function being executed, if you do not delete it from the configuration for the line skip procedure.

Another big disadvantage is that the skipped lines could throw errors, when they get parsed.
My intentional thought of this feature was, that, no matter what kind of input is in the first few lines being skipped, the parser will not give any error messages due to the fact, that the first few lines might even be just some text or information in another format but csv.
Maybe the feature could be named _dataStartsAtLine_ then too.

I thought of the feature from this myTracks extract:

	"Name","Art der Aktivität","Beschreibung"
	"Pallasstraße 35","Fahrrad fahren",""

	"Segment","Punkt","Breitengrad (°)","Längengrad (°)","Höhe (m)","Peilung (°)","Genauigkeit (m)","Geschwindigkeit (m/s)","Zeit","Leistung (W)","Trittfrequenz (Umdrehungen pro Minute)","Puls (Herzschläge pro Minute)"
	"1","1","52.494246","13.360865","90.5999984741211","","49","","2015-02-17T17:44:01.158Z","","",""
	"1","2","52.494269","13.360882","90.5","","45","","2015-02-17T17:46:29.753Z","","",""
	"1","3","52.494337","13.36094","90.69999694824219","","47","","2015-02-17T17:46:32.948Z","","",""

In this case it's two different sets of csv data in one file.

But what about this:

	Here we just describe what the following csv data is about.
	Now just think what happens with this following "line
	break" it will be skipped with the latest code i provided
	because we assume it's csv-like data...
	"Segment","Punkt","Breitengrad (°)","Längengrad (°)","Höhe (m)","Peilung (°)","Genauigkeit (m)","Geschwindigkeit (m/s)","Zeit","Leistung (W)","Trittfrequenz (Umdrehungen pro Minute)","Puls (Herzschläge pro Minute)"
	"1","1","52.494246","13.360865","90.5999984741211","","49","","2015-02-17T17:44:01.158Z","","",""

3. What should the feature be about?
---
Now i would suggest to decide on if this feature is really something to be added and if it is going to be added, what it should aim at:
* Skipping lines of data with __no__ data type assumption (__My actual feature request__; if there is a newline it just gets processed leaving out any thought about double quotes or the likes)
* Skipping actual lines of csv data
	* we would have to consider headings; i did not do this in my provided code (and this is also where utilizing the preview code might come in handy)
	* __but__: i can't think of a logical reason to skip actual data; i think a user would get rid of unwanted data by filtering it after being parsed

My latest code is right in between these two thoughts now as it doesn't parse csv properly (ignoring any formatting in between a first double quote and a double quote following either the delimiter or a newline - __it is actually not working properly atm__) nor can it process my second example given above rendering the code useless.
My first code just assumed the first thought; but i can sense it's not the smartest to just look out for _newlines_ as it is pretty error prone, when the data at the beginning of a file does not follow a strict structure. Either you offer the feature with some restrictions (skipped lines do not get parsed, i.e. newlines are newlines no matter what) or you stick to suggest using __comments__ for things like this.

I'm very undecided about this at the moment. I'd like to hear your thoughts about this.

bluej100:
Perhaps we could add a callback which allowed users to manipulate chunks
before parsing begins? I think that would support all of the use cases you
describe here.

zemzelett:
I had a similar idea yesterday. Something like a preprocessor.
I'll think about this in the course of next week.

bluej100:
I've tagged this "discussion", as the current patch is probably not the desired behavior.

mholt:
So did we decide that beforeFirstChunk was a more fitting way to solve this problem? If so, I'm in favor of closing this and the related issue. :+1:

PR #192, opened 2015-03-27
beforeFirstChunk callback, fixes #186

### LINKED ISSUE #186: Skip lines at the beginning of a file? (Feature Suggestion)

bluej100:


papaparse.js
```diff
@@ -389,6 +390,10 @@
 
 		this.parseChunk = function(chunk)
 		{
+			if (this._chunkIndex == 0 && isFunction(this._config.beforeFirstChunk))
+				chunk = this._config.beforeFirstChunk(chunk);
```
mholt: In case beforeFirstChunk doesn't return anything, we should still use the chunk as-is. Something like `|| chunk` at the end of the line (except maybe compare against `undefined` so we acknowledge empty strings)

mholt:
Thanks for all the awesome tests, btw!

papaparse.js
bluej100: Done.

mholt:
Awesome - the only thing to note about this change is that the new `_chunkIndex` variable isn't necessarily the index of the chunk - it's the number of times `parseChunk()` was called, including from `resume()` which operates on the same chunk (more or less).

If that's not a concern, I can merge this in. I might rename the variable later but looks good to me otherwise!

bluej100:
Good point. I'm hoping to write a new patch for https://github.com/mholt/PapaParse/issues/164 which removes the "fake chunks", happily.

mholt:
Will push a changed variable name in just a few minutes. Thanks!

PR #194, opened 2015-03-30
Removed result variable for file parsing.

cmfcmf:
As it says, file parsing happens asynchronously, so there is no need to safe to the results variable.


mholt:
Good catch. Thanks! I wonder how many people that line confused...

cmfcmf:
Well, at least me :smirk:

PR #197, opened 2015-04-03
FieldMismatch Error row error fix

zemzelett:
Hi there,

as i already stated last week that i ran against an error while processing a file in small chunks. When there are several chunks, the row number displayed in the errors object is not correct as it is referring to the line processed in the current chunk ignoring all chunks (respectively rows) that were processed before.

The fix is simple and shouldn't interfere with anything else. Or am i missing something?


tests/test-cases.js
```diff
@@ -1319,5 +1319,20 @@ var CUSTOM_TESTS = [
 				}
 			});
 		}
-	}
+	},
+	{
+		description: "Errors give correct row numbers with several chunks being processed",
+		expected: '[{"type":"FieldMismatch","code":"TooFewFields","message":"Too few fields: expected 12 fields but parsed 11","row":21},{"type":"FieldMismatch","code":"TooFewFields","message":"Too few fields: expected 12 fields but parsed 10","row":23}]',
```
bluej100: It looks to me like these are referring to line 23 and 25 in the file. Shouldn't it be row 22 and 24?

papaparse.js
```diff
@@ -850,9 +850,9 @@
 				{
 					_results.data[i] = row;
 					if (j > _fields.length)
-						addError("FieldMismatch", "TooManyFields", "Too many fields: expected " + _fields.length + " fields but parsed " + j, i);
+						addError("FieldMismatch", "TooManyFields", "Too many fields: expected " + _fields.length + " fields but parsed " + j, self.streamer._rowCount + i);
 					else if (j < _fields.length)
-						addError("FieldMismatch", "TooFewFields", "Too few fields: expected " + _fields.length + " fields but parsed " + j, i);
+						addError("FieldMismatch", "TooFewFields", "Too few fields: expected " + _fields.length + " fields but parsed " + j, self.streamer._rowCount + i);
```
bluej100: Perhaps we could do this in addError instead of fixing it in every occurrence? Also, "MissingQuotes" has the same problem.

tests/test-cases.js
zemzelett: Well yes, the errors are in line 23 and 25, but considering 0-base of arrays and the headers you get the resulting rows of 21 and 23. Without the headers your assumption would be correct.

papaparse.js
zemzelett: You're right, it's possible to add this in addError... like this:

`row: row ? self.streamer._rowCount + row : row`

"MissingQuotes" is a little harder to fix. I can see an option in handing _rowCount to the _parser (like: `_parser.rowCount = self.streamer._rowCount;`) but this would probably end up in some core parser tests failing, as the variable cannot get handed in there working directly on the core parser. As long a you don't test for it not being defined like: `row: (this.rowCount ? this.rowCount : 0) + data.length,	// row has yet to be inserted`
Seems a little off to me.

mholt: What if the error is actually in row 0? Would that still work?

zemzelett: First i thought yes... but you mean when it's in row 0 of a chunk, right? I'm gonna check that.

@mholt @bluej100 Apart from that, what do you think of touching the main parser like that (for fixing MissingQuotes row error)? I still think it's not the best way, to do that. Can you think of some other to do it?

zemzelett: @mholt It wouldn't, the line should be changed to something like this:
`row: (row !== undefined) ? self.streamer._rowCount + row : row`

I'm going to run some test cases and give feedback as soon as i can.

mholt:
@zemzelett What's your status on this?

pokoli:
I'm closing this due as it have been worked in the last two years. Feel free to reopen if you want to work on it.

Issue #198, opened 2015-04-14
Papa.unparse(Papa.parse(foo))


pykiss:
It would be nice can to can parse a file, modify the parse result and unparse back easily.

For instance, 

    var csv = Papa.parse(foo, {header:true});
    detele csv.meta.fields[0];
    return Papa.unparse(csv);

pykiss:
Ok, I just see that it is really easy.

   var csv = Papa.parse(foo, {header:true});
   csv.fields = csv.meta.fields;
   detele csv.meta.fields[0];
   return Papa.unparse(csv);

mholt:
Making the parse output and unparse input symmetric would be really useful, actually. I'd you don't mind I will reopen this so I can remember to take a closer look to see if that's feasible.

PR #199, opened 2015-04-14
symetric parse and unparse

pykiss:
(I have not test that, sorry)

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -248,7 +248,7 @@
 			if (_input.data instanceof Array)
 			{
 				if (!_input.fields)
-					_input.fields = _input.data[0] instanceof Array
+					_input.fields = ( _input.meta && _input.meta.fields ) || _input.data[0] instanceof Array
 									? _input.fields
 									: objectKeys(_input.data[0]);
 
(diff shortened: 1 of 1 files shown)


mholt:
Awesome, thanks for taking the initiative!

Final exams start this week. I'll write a test for this afterward unless somebody beats me to it.

mholt:
[ Connecting this with related issue, #198 ]

papaparse.js
```diff
@@ -248,7 +248,7 @@
 			if (_input.data instanceof Array)
 			{
 				if (!_input.fields)
-					_input.fields = _input.data[0] instanceof Array
+					_input.fields = ( _input.meta && _input.meta.fields ) || _input.data[0] instanceof Array
 									? _input.fields
```
mholt: This needs a little work; if the input is coming from Papa.parse(), then the fields are in `_input.meta.fields`. But we still need to account for the case when the fields are defined in `_input.fields` too. Might require turning this ternary statement into a full `if` and `else` statement or something, to make it easier to read.

mholt:
@pykiss Thanks. We still need tests for this. Would you be able to write one?

Issue #140, opened 2014-12-14
node compatibility


bluej100:
It would be nice to have node compatibility as part of this project, including streamers for Node resources.

I think I'll work on this after the Grunt build and streamer refactoring pull requests merge.

robd:
I just wanted to update this issue with the progress we've made. As discussed over in #167, I think PapaParse should support basic parsing etc with Node now that #172 has been merged. It might be good to update the PapaParse docs to indicate that there is now some Node support.

There is more cleanup possible in this area, for example, to improve the separation of web dependent features such as web workers, remote file parsing, file object support, multithreading etc, so that they are never loaded / supported when executed in a Node environment. We could also look at Node ports of some of these if they make sense. I'm not sure what the use cases are there, but it might be sensible to prioritize the possible Node features based on what's needed.

Another question is whether BabyParse is still required. AFAICT BabyParse doesn't include any tests for behaviour on Node, so it's a bit difficult to know what functionality is or isn't supported there, but I think it's possible that PapaParse now supports everything that BabyParse does. Perhaps @Rich-Harris could see if PapaParse would now work for his application. 

From a maintenance perspective, and for newcomers to PapaParse, I think it would be more straightforward not to have 2 projects (ie Baby and Papa) both supporting node, so it might make sense to sunset BabyParse and pick up Node support in Papa directly.

mholt:
I would like to sunset Baby Parse as soon as Papa Parse can run in Node. Like you said -- and majorly thanks to you and Braden -- we are getting closer to that point.

While I don't think we need to implement multi-threading for Node environments, it might be handy to stream and download files using some native Node facilities. Node compatibility is something that I will not personally implement with Papa Parse. (I don't have Node installed because I don't use Node, and I have no personal interest in developing for it at this time.) However, I support anyone who is willing to make it happen.

First, let's make what we already have play nicely in Node by making sure the browser-only features stay out of the way. Once Papa Parse can run out-of-the-box in a Node environment, I will make a release, after which we can consider adding bells and whistles.

Thanks for all your contributions thus far!

robd:
OK, cool. I'm also not using node, but if I have time I will have a crack at better separating the web specific stuff somehow. 

I think the latest version PP will already work with Node without further modifications needed by the user. So could you expand on what you're thinking when you say 'out-of-the-box' in a Node env?

Thanks!

mholt:
[...]

PR #201, opened 2015-04-23
node compatibility #140 - Add file parsing support

bmcorum:
Supports CSV files with either a csv or txt extension. Supports single file or array of files. When passing an array of files you can pass in an override config ( { file: filename, config: overrideConfig } ).

Examples

parseFiles(file1.csv)

config = { ... }
parseFiles(file1.csv, config)

parseFiles([file1.csv, file2.csv], config)

overrideConfig = { ... }
parseFiles([{ file: file1.csv, config: overrideConfig}, file2.csv], config)


papaparse.js
```diff
@@ -167,6 +168,42 @@
...
+			});
+			return results;
+		} else {
+			var results = {
+				data: [],
+				errors: []
+			};
+			if ((/(\.csv|\.txt)$/).test(_input)) {
+				try {
+					var contents = fs.readFileSync(_input).toString();
```
mholt: How well does this work with really large files?

papaparse.js
```diff
@@ -167,6 +168,42 @@
...
+				else
+					results.push(ParseFiles(input, _config));
+			});
+			return results;
+		} else {
+			var results = {
+				data: [],
+				errors: []
+			};
+			if ((/(\.csv|\.txt)$/).test(_input)) {
```
mholt: Let's not restrict the file extension - there could be any number of different valid ones, or none at all.

papaparse.js
```diff
@@ -167,6 +168,42 @@
...
+					return results;
+				}
+			} else {
+				results.errors.push({
+					type: '',
+					code: '',
+					message: 'Unsupported file type.',
+					row: ''
+				});
+				return results;
```
mholt: @bmcorum Again, I don't think we need to worry about this. But good for thinking of it :+1:

papaparse.js
bluej100: Yeah, readFileSync seems like the wrong call here (see http://stackoverflow.com/questions/13822085/fs-readfilesync-seems-faster-than-fs-readfile-is-it-ok-to-use-for-a-web-app-in for perspective from someone who does more Node than I do). I would suggest that you write a ChunkStreamer subclass (like FileStreamer) so you get the benefit of all of Papa's chunked reads and async processing.

bmcorum:
@bluej100  I would need to wrap the ParseFiles in a promise which would not be terribly difficult. I will need to find the time to write and test the code.

@mholt Not sure how large a "large" file would be but I have a few that are a few hundred lines long (are we talking 100s of KB, 100s of MB, or ***I hope not 100s of GB). I am currently using Baby since Papa is a bit bloated for strict Node use but I can understand using Papa when trying to target multiple platforms.

pokoli:
I'm closing this in favour of #370

PR #204, opened 2015-05-04
avoid division by zero

andekande:
if preview didn't find a row, avgFieldCount will be NaN or Infinity

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -903,7 +903,8 @@
 					}
 				}
 
-				avgFieldCount /= preview.data.length;
+				if (preview.data.length > 0)
+					avgFieldCount /= preview.data.length;
 
 				if ((typeof bestDelta === 'undefined' || delta < bestDelta)
 					&& avgFieldCount > 1.99)


mholt:
Hm, interesting scenario! Code looks sound, at a glance. So you tested this and it works? (I'm headed out the door so I just thought I'd ask.)

andekande:
i have been running tests and they were green - no side effects introduced

mholt:
Can you add a new test that causes a division by 0? That way we an assert that we don't end up breaking this accidentally later on down the road.

andekande:
The PARSE_TESTS "Empty input string" and "Skip empty lines, with empty input" already causes this.

mholt:
@andekande I still don't see it. Is there anything else required for reproducing it?

screen shot 2015-06-23 at 11 21 38 am

andekande:
Well, there currently is not any fault existent. But there potentially is. Consider this:
```javascript
avgFieldCount /= preview.data.length;
 
if ((typeof bestDelta === 'undefined' || delta < bestDelta) && avgFieldCount > 1.99)
```
avgFieldCount can get "Infinity" which always is > 1.99
but this if is never hit because of the additinal bestDelta check.

The Fix is only a measure of precaution.

mholt:
Got it. Thanks!

PR #205, opened 2015-05-06
Option for replacing smart quotes with ASCII quotes

wjlroe:
This adds an option that enables replacing smart quotes with ASCII quotes that will be understood by PapaParse. It's not on by default because (as it currently is) it requires scanning the entire string, replacing a regex.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -972,6 +972,7 @@
 		var step = config.step;
 		var preview = config.preview;
 		var fastMode = config.fastMode;
+		var replaceSmartQuotes = config.replaceSmartQuotes || false;
 
 		// Delimiter must be valid
 		if (typeof delim !== 'string'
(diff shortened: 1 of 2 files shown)


bluej100:
This would be one case where a beforeChunk (rather than just beforeFirstChunk) would be useful.

mholt:
@bluej100 That's true. I would like to keep as much stuff out of the core parser as possible.

mholt:
@wjlroe Is it possible to fix the file before submitting it for parsing?

mholt:
@wjlroe @bluej100 I think I'd rather implement a beforeChunk callback, if it proves feasible. My concern is its performance in a worker. The chunk will have to be copied to the main thread, then copied back to the worker before processing that chunk even takes place. (For that matter, does `beforeFirstChunk` work with workers? I'd check now but I'm on a roll going through these PRs and issues.)

bluej100:
You certainly are on a roll. :pray: 

`beforeFirstChunk` does not currently work with workers, no. I agree with your performance/complexity concerns; in my opinion, it would be reasonable to spit out an error to people who specify both `beforeChunk` and `worker`, since developers will almost certainly see that before it hits production. I think it's pretty reasonable to expect people to write their own worker if they're doing fancy stuff, even though I do think it's cool that Papa has built-in worker handling for the simple case.

pokoli:
I'm closing this PR as I understand the proposed behaviour is out of the scope of the parser. Re-open it if you don't agree.

Issue #163, opened 2015-01-21
When use worker, I get Uncaught TypeError: Cannot read property 'defaultView' of undefined


yokeholy:
I'm using 4.1.0 minified version. Everything works fine until I turn on worker.
No matter which way I call Papa Parse (Papa.parse(file,config) or $(file).parse(config)), I will get this error in my console and then nothing happens (the parser stops).

mholt:
What kind of environment are you running Papa in? I've seen that reported before; it's probably the same as #114 but we weren't able to pin it down.

I have a hunch now, though. Papa Parse explicitly forbids the use of the byte order mark (`\ufeff`) as a delimiter, so this character is defined as a variable for convenience (in the minified version, it's `e.Papa.BYTE_ORDER_MARK`). I just noticed that the minified version seems to replace the sequence `\ufeff` with the literal byte order mark. Maybe that's problematic.

Could you take a look at position 12,708 in papaparse.min.js - there's the unescaped unicode character there. @tony-cocco, you also had this issue, if you could take a look too I would appreciate it. Try removing that line and the reference to it in the BAD_DELIMITERS array.

If you take out the byte order mark, does it work?

tony-cocco:
I can't say definitively, but when I stopped passing the papaparse library through minification my issue went away. I would suggest @yokeholy try using the un-minified library and see if that solves his problem.

If possible, might be worthwhile to know what the file that's being parsed looks like.

yokeholy:
I tried both minified and un-minified versions. The problem exists whenever worker is on.
I tested with large files (90MB+) as well as a single line string. Quoted strings with comma delimiter.
My environment is Angular+jQuery+BootstrapJS.

yokeholy:
One thing through my research pops up: http://stackoverflow.com/questions/3936211/jquery-cannot-read-property-defaultview-of-undefined-error

It seems like the scope has changed. I wonder if it's anything to do with Angular because I'm calling the parser within an Angular controller scope.

tony-cocco:
I was also using Angular as well. But, the scope changing could still be the cause. My issue came from the grunt task messing with the papa library though.

Apparently, I was using version 3.1.3 of papa. Sorry I can't be of more help.

yokeholy:
So I did another test and put the parser caller function outside of my Angular scope, unfortunately, the problem still exists.

yokeholy:
@mholt I looked into position 12708, I didn't find the unescaped unicode character there.
[...]

PR #210, opened 2015-05-22
More visible explanation of concatinating issue (#163) in the docs

nathancahill:
I read the docs thoroughly but missed the note in the middle of the FAQ about concatenating Papa Parse with other libraries. This adds that (pretty significant) caveat to the `config.worker` description and the Multi-Threading (Workers) FAQ since concatenating JS for production is pretty standard these days.


mholt:
Thanks for improving the docs!

PR #211, opened 2015-05-22
Using PapaParse inside a worker

### LINKED ISSUE #188: Using PapaParse inside a worker
I have the problem that the parser always wants to post the results using postMessage because I'm using PapaParse inside a worker.  I guess it's because it thinks it is one of the workers the parser spawned. 

Is there any kind of global where I can deactivate all the worker behaviour?

benekastah:
Fixes #188

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -7,7 +7,9 @@
 {
 	"use strict";
 
-	var IS_WORKER = (!global.document && !!global.postMessage), LOADED_SYNC = false, AUTO_SCRIPT_PATH;
+	var IS_WORKER = !global.document && !!global.postMessage,
+		IS_PAPA_WORKER = IS_WORKER && /(\?|&)papaworker(=|&|$)/.test(global.location.search),
+		LOADED_SYNC = false, AUTO_SCRIPT_PATH;
 	var workers = {}, workerIdCounter = 0;
 
 	var Papa = {};
@@ -19,7 +21,7 @@
 	Papa.UNIT_SEP = String.fromCharCode(31);
 	Papa.BYTE_ORDER_MARK = "\ufeff";
 	Papa.BAD_DELIMITERS = ["\r", "\n", "\"", Papa.BYTE_ORDER_MARK];
-	Papa.WORKERS_SUPPORTED = !!global.Worker;
+	Papa.WORKERS_SUPPORTED = !IS_WORKER && !!global.Worker;
 	Papa.SCRIPT_PATH = null;	// Must be set by your code if you use workers and this lib is loaded asynchronously
 
 	// Configurable chunk sizes for local and remote files, respectively
@@ -145,7 +147,7 @@
 	}
 
 
-	if (IS_WORKER)
+	if (IS_PAPA_WORKER)
 	{
 		global.onmessage = workerThreadReceivedMessage;
 	}
@@ -426,7 +428,7 @@
 
 			var finishedIncludingPreview = this._finished || (this._config.preview && this._rowCount >= this._config.preview);
 
-			if (IS_WORKER)
+			if (IS_PAPA_WORKER)
 			{
 				global.postMessage({
 					results: results,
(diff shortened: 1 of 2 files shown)


mholt:
Cool, I'll take a look at this soon!

benekastah:
@mholt If you can look at this soon, I'll fix the conflicts.

mholt:
@benekastah Thanks for the push :smile: Looked at it just now. Looks fine, seems like a logical way to do it. Haven't tried it but if the tests pass, let's merge it in!

benekastah:
Cool, I'll fix up the conflicts and make sure the tests pass then.

bluej100:
I think we might want to separate IS_WORKER into two separate flags: one for whether we are in a worker and one for whether the current parse was called directly or through a posted message. For example, the IS_WORKER checks for synchronous XMLHttpRequest and FileReader usage should probably continue to be based on the !global.document check alone. My suggestion would be that instead in workerThreadReceivedMessage we add an isWorker property to the config and check that in chunkError, etc.

bluej100:
Certainly it's unlikely that a single worker would do parses both directly and indirectly, so the second flag could reasonably be a constant as well.

benekastah:
Ok, I should have fixed the tests. @bluej100 I think I know what you mean, I'll make that modification now and see if it's what you had in mind.

mholt:
@bluej100 As you know, it's Pioneer Day here in Utah so I won't be online most of today. If you like what you see, feel free to merge the PR when it's ready.

bluej100:
Sir.

benekastah:
@bluej100 Ok, what I did is slightly different than what you described, but I think it captures the spirit of what you wanted. I couldn't just add an `isWorker` property to the config in `workerThreadReceivedMessage` because the main problem I was trying to prevent was that papaparse was setting `workerThreadReceivedMessage` as the `onmessage` callback even if papaparse wasn't the worker's owner. So I changed the query string to be "papaworker" rather than just "worker" as the signal the papaparse spawned it and is the owner.

tests/test-cases.js
```diff
@@ -1458,6 +1458,49 @@ var CUSTOM_TESTS = [
 				}
 			});
 		}
+	},
+	{
+		description: "Should not assume we own the worker unless papaworker is in the search string",
+		expected: [false, true, true, true, true],
+		run: function(callback) {
+			if (typeof Worker === 'undefined') {
```
bluej100: Perhaps you could use the disabled property here, as used with FILES_ENABLED?

papaparse.js
```diff
@@ -1395,3 +1400,5 @@
 		return typeof func === 'function';
 	}
 })(typeof window !== 'undefined' ? window : this);
+
+// vim: noet
```
bluej100: Mind removing this?

tests/test-cases.js
```diff
@@ -1458,6 +1458,49 @@ var CUSTOM_TESTS = [
...
+				}
+			};
+
+			searchStrings.forEach(function (searchString, idx) {
+				var w = new Worker('../papaparse.js' + searchString);
+				// Give it .5s to do something
+				var timeout = setTimeout(function () {
+					w.terminate();
+					setResult(idx, false);
+				}, 500);
```
bluej100: You could simplify this a bit by initializing to false and having a single timeout with callback(results), if you like.

bluej100:
Looks good to me. Left a couple suggestions on the test. Thanks for your work on this, @benekastah .

benekastah:
Cool. Addressed all your comments.

benekastah:
I changed `WORKERS_SUPPORTED` per your suggestion (but I don't find your comment anymore...)

bluej100:
Merged. Thanks a bunch!

benekastah:
Thank you!

benekastah:
Is there a way I can get notified when this PR makes it into a new version?

mholt:
As soon as I can get around to it, I'll try to remember. Been really busy with other projects and travel.

mholt:
@benekastah Deployed in version 4.1.2.

benekastah:
Thanks @mholt!

PR #212, opened 2015-06-03
Added support for ignoring opening quotes

r1c4r60:
Added support for ignoring opening quotes

New configuration: openingQuotes ::: Boolean
Default === true

When openingQuotes === true, opening quotes are processed and errors are
thrown when not terminated properly.
i.e.
'A,"B" C,D,E' ->
errors: [
{
"type": "Quotes",
"code": "MissingQuotes",
"message": "Quoted field unterminated",
"row": 0,
"index": 3
}
]

When openingQuotes === false, quotes are ignored.
i.e.
'A,"B" C,D,E' -> [['A','"B" C','D','E']]


bluej100:
Do your files have no support for fields containing commas, then?

mholt:
The CSV format requires that quotes in the data are contained in quoted fields and escaped properly. Is there a reason your input cannot conform to the CSV spec?

r1c4r60:
### 1. *bluej100* 
> Do your files have no support for fields containing commas, then?

The problem I faced was not with commas, but with quotes. 

### 2. *mholt* 
> The CSV format requires that quotes in the data are contained in quoted fields and escaped
> properly. Is there a reason your input cannot conform to the CSV spec?

Data used: Copy & pasted data from Excel worksheet into a '.csv' file. It results in a tab delimited dataset per column. The data is a large amount of open-ended survey replies where respondents at times used quotes in their responses. i.e. The place was "cozy". The decorations was "modern".  

The options in this case were: 
 * Either to escape a pretty large data set and parse with i.e. PapaParse.
or
 * Parse with PapaParse using one elegant configuration even though it conflicts with the CSV "standard" for reserved characters. Noticed the double quotes !? :)

#### I chose the latest. No harm done if you don't think this pull request is suitable. 

**Pros of accepting this modification:** 

1. The normal parsing speed of PapaParse will increase since, in runtime, you wouldn't have to verify the cell's first character for an opening quote when the new configuration would be set to false. 
2. Allows for one more use case where the user is sure of the content of each cell and just wants the data to be split.

** Negatives **

1. The normal parsing speed of PapaParse will **decrease** slightly (guessing that nearly imperceptibly) since, in runtime, you would have to verify the the new configuration variable AND the cell's first character. This could be resolved easily though.
2. Disregard for the CSV standard.

bluej100:
> The problem I faced was not with commas, but with quotes. 

Ohhh, because your file is tab-delimited. Well, tab characters do seem less likely to be present inside cells.

mholt:
I do not recommend generating CSV files by copying and pasting out of spreadsheet programs. Why not just use their File -> Export functionality? That will do it correctly.

mholt:
@r1c4r60 Does File -> Export work for you? If so, I think we can close this.

r1c4r60:
@mholt I am using my forked version and life is good. Thanks for the consideration! Have a great week!

PR #216, opened 2015-06-04
error when line ends with doublequotes

andekande:
simplest Testinput: ""a""\n""b""\n""c""
the cause ist to be found in Line 1099 of the papa source

### DIFF
--- a/tests/test-cases.js
+++ b/tests/test-cases.js
@@ -206,6 +206,14 @@ var CORE_PARSER_TESTS = [
 			errors: []
 		}
 	},
+    	{
+		description: "Line ends with double quoted field",
+		input: 'a,b\n"c","d"\n""e"",""f""\n""g"1",""h"1"\n"1"i"","1"j""\n"1"k"1","1"l"1"',
+		expected: {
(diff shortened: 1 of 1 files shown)


mholt:
I think this syntax is not doing what you expect:

```
""e"",""f""
```

This is invalid, since the first quotes opens the quoted field, and the next adjacent quotes need to be escaped with yet another quotes mark. So you probably are looking for this, instead:

```
"""e""","""f"""
```

It's a ridiculous amount of quotes, but that results in the field data being `"e"` (with quotes in the data).

andekande:
jup, you are right - the standard wants it with double double quotes
Still, since PP want's to be "robust" it might tackle those "errors" ... ?

mholt:
Well, robust != lenient. Robust means it handles errors and *edge cases* gracefully and properly. Edge cases, however, are not breaking the rules - just on the edge of them. I'm going to choose the conservative option here, and opt to not change the quotes logic. Quotes are complicated and a huge reason that CSV parsing is slow/hard, and I'd rather not coding in exceptions.

Thanks for your report, though! This is an important thing to understand about Papa Parse.

PR #224, opened 2015-06-25
Updated double equals to triple equals, removed whitespace at end of lines and standardized quotes.

breck7:
Just ran the code through a linter and made some code style changes. All tests pass. Perf difference of == vs === is negligible IIRC but jsperf is currently down so don't have data at the moment.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -532,6 +532,10 @@
 
 			xhr = new XMLHttpRequest();
 			
+			if (this._config.withCredentials)
+			{
+				xhr.withCredentials = this._config.withCredentials;
+			}
 			if (!IS_WORKER)
 			{
 				xhr.onload = bindFunction(this._chunkLoaded, this);


breck7:
No need to accept if you don't think this makes a difference. These are just nits :)

mholt:
Cool, thanks. Will have to wait for jsperf to come back online to be sure about performance though. Stay tuned. :smile:

breck7:
If the types are different === is signficantly faster, otherwise they are about equal. http://jsperf.com/triple-equals-vs-double-equals/. So this could be faster.

mholt:
@breck7 Thanks for the info. Unfortunately it doesn't seem to be all the way online yet, all I see is:

> The get/post request has failed :(

mholt:
@breck7 When you get a chance, sync up with master and I'll merge it in :+1:

PR #230, opened 2015-07-06
Add version for bower package install

danblundell:


mholt:
Thanks, but [bower ignores this and uses git tags instead](http://bower.io/docs/creating-packages/#version).

danblundell:
Only added as bower's not letting me install because there's no version defined...

mholt:
That sounds like a bower issue. Try asking them first then let me know if this is required now.

PR #239, opened 2015-07-23
handle empty quoted field at EOF

### LINKED ISSUE #238: Incorrect result when last record ends with an empty quoted value
Consider this example:
```csv
a,1,""
b,2,""
c,3,""
```
Parsing this structure results in the following data:
```json
[
    [
        "a",
        "1",
        ""
    ],
    [
        "b",
        "2",
        ""
    ],
    [
        "c",
        "3",
        "\""
    ]
]
```
Notice that the last string value contains an extra double quote, which is incorrect IMHO. I'd at least expect every record to be treated equally.

bluej100:
fixes #238

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -1201,7 +1201,7 @@
 			{
 				if (ignoreLastRow)
 					return returnable();
-				if (!value)
+				if (typeof value === 'undefined')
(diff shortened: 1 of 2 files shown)


mholt:
Ah, thank you!

PR #277, opened 2015-11-18
Fix newline option for delimiter autodetect

jbzorg:
In function guessDelimiter always used '\n' as newline. This pull request fix it.


mholt:
Ah, thank you.

Issue #285, opened 2016-02-06
Parser chokes on extra EOF linebreak


jeroenooms:
I have a somewhat strangely formatted CSV file like [this](http://web.ohmage.org/~snolen/sample_classes/01sample_class.csv). The file uses `\r\n` linebreaks and has an additional linebreak at the end of the file. When parsing with `{header : true, skipEmptyLines : true}`

```
[row 11] Too few fields: expected 20 fields but parsed 1
[row 12] Too few fields: expected 20 fields but parsed 1
``` 

The problem goes away if I remove the EOF linebreaks, but unfortunately the csv input is beyond my control. Shoulnt `skipEmptyLines` be skipping redundant linebreaks?

mholt:
What happens if you set `newline: "\r\n"`in the config?

jeroenooms:
It fails with an error from the trailing empty line

mholt:
More info, please -- what's the error? And the file has several trailing empty lines, which one?

jeroenooms:
If I parse with:

```js
Papa.parse(file,{
	header : true,
	skipEmptyLines : true,
	newline: '\r\n',
```

I get:

     [row 11] Too few fields: expected 20 fields but parsed 1

mholt:
Does the error only appear once? i.e. do you get an error on row 12?

jeroenooms:
Yes only once. You can easily reproduce it with the [csv file](http://web.ohmage.org/~snolen/sample_classes/01sample_class.csv)

mholt:
(Closing for reasons explained in #286)

PR #286, opened 2016-02-09
Trim empty lines from whitespace

jeroenooms:
Fixes https://github.com/mholt/PapaParse/issues/285

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -816,7 +816,7 @@
 			if (_config.skipEmptyLines)
 			{
 				for (var i = 0; i < _results.data.length; i++)
-					if (_results.data[i].length === 1 && _results.data[i][0] === '')
+					if (_results.data[i].length === 1 && _results.data[i][0].trim() === '')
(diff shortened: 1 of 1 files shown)


mholt:
There are tests to assert against this behavior which are now failing; I do not think this is what is intended, since technically the line is not empty.

jeroenooms:
Do you have another suggestion to fix https://github.com/mholt/PapaParse/issues/285 ?

mholt:
I don't think it's broken. Whitespace is not an empty line.

jeroenooms:
Is there any way to parse a file like this then? http://web.ohmage.org/~snolen/sample_classes/01sample_class.csv It has `\r\n` linebreaks with an extra EOF. Unfortunately this input format is beyond my control.

mholt:
You'll have to just filter those out after you get the results back, I think. Papa is a parser, not a data cleanser. Thanks for the PR though!

Issue #287, opened 2016-02-09
Trim header column names


slgal:
Idea is a trim spaces at the begin/end of columns name.

Example:
Product Name,SKU,Product Code, Price ,Member Price, Cost,Stock Quantity

So price header will be 'Price' insted of ' Price ' and cost will be 'Cost' insted of ' Cost'

mholt:
I'd rather Papa just parse the data as-is and let the user clean it up.

larryboymi:
Sometimes with non-printable characters this becomes pretty unintuitive.

pokoli:
@larryboymi see #589

larryboymi:
Oh thanks! I didn't see that in the documentation, but now I do. Much appreciated.

mtmacdonald:
In case it helps anyone else, 4x version does have `trimHeaders` option (undocumented). And 5x version has `transformHeader`.

PR #288, opened 2016-02-09
#287

slgal:


mholt:
Papa shouldn't be changing the data that it parses. Thanks though!

slgal:
Understand.
Thank you for the great tools.
9 лют. 2016 17:38 "Matt Holt" <notifications@github.com> пише:

PR #301, opened 2016-04-02
Fix for complete(res, file) in parse(file, ...)

AstroAndyN:
Fix the "complete(res, file)" for parse(file, ...) so that "file" is not undefined in "complete". Works for the tested case, but it is not clear if this might cause problems with other "Streamed" reading modes.


mholt:
Thank you!

PR #306, opened 2016-04-19
Fix undefined global variable in UMD module

IceOnFire:
See http://stackoverflow.com/questions/3277182/how-to-get-the-global-object-in-javascript for reference.


IceOnFire:
I re-forked the project and custom tests fail; why does Travis always says they're ok?

Please wait before merging this PR.

PR #308, opened 2016-04-19
Fix global behaviour in UMD module

IceOnFire:
There were a couple of errors preventing PapaParse from being used as a global variable. One is a mere copy-and-paste slip, the other is a common issue described [here](http://stackoverflow.com/questions/23834572/global-context-inside-umd-pattern).

Now all tests pass, I'm quite confident this will work for everyone.


mholt:
Thank you! :smile: Really appreciate it.

PR #315, opened 2016-06-11
Support pipe and multi-core?

aeromomo:
I am trying to use this to parse large chunks of data. Can you add support for pipe and multi-core?


mholt:
This is a pull request for a different project.

Issue #218, opened 2015-06-04
Doublequotes vs Singlequotes


andekande:
I'd like to suggest that the Quote Character can be swappable. There might be CSV Files which use ' instead of " to Quote their Fields. All Rules (escaping, closing, etc) still apply. So we would just need to replace the hardcoded Doublequotes with a Variable.
I already have a prototype running. Question is whos burden it is to detect the true Quote Character. My simple "guessing" approach now: 
````
var input = "'a','b'\n";
var i2 = input.trim();
if (i2[0] == "'" && i2[i2.length-1] == "'")
    quoteChar = "'";
````

Or one knows how the Files are formatted and might supply a config param, just like with newLines.

Topic open for Discussion...

andekande:
Here is my current take on it which also fixes the return value for chunked strings beeing an empty array (Papa.parse("", {chunkSize: 100});):
https://github.com/andekande/PapaParse/commit/1345bfa0210919bc8e0c23b35469dd4b46af3c9e#diff-eb20bdfe95a768f4eecdf3bebb522a4e

mholt:
After looking into this a bit, I have a few thoughts/concerns.

RFC 4180 requires that CSV quoted fields be contained specifically in *double quotes*. I hesitate to deviate from this standard much, especially on the matter of quotes which are the whole reason that CSV parsing is difficult. I would recommend that single quotes around fields be replaced with double quotes by the user before sending it to Papa Parse.

If it were to be done, auto-detecting the quotes character is probably very difficult to get right, especially with certain edge cases. I shudder at the thought. :fearful:

The reason that newlines are customizable is because, regrettably, different systems use different characters, and it's so very common that CSV files only have LF to separate lines, even though technically CRLF is required. But because this was so common and the characters are not easily accessible from the keyboard, we had to handle the different newlines. In contrast, I don't see any reason why a user couldn't just use the proper double quotes character. (If I had my way, ASCII 30 would be on the keyboard and CSV would become a delightful format to use!)

Also, any superfluous changes to the core parser like this one must be at least as fast as the current implementation.

bluej100:
+1 no autodetection, at least

andekande:
Well it's not too uncommon that the quotechar is not a doublequote, in the end RFC acts more like a suggestion in terms of CSV...
There are Libraries that support switching the QuoteCharacter, e.g.:
Standard CSV Parser in Python: https://docs.python.org/3.3/library/csv.html
[...]

PR #317, opened 2016-06-16
Add custom quote character option for parsing CSV files

prayashm:
Note: Some characters like * or ^ don't work properly as custom quote character.
Kindly check the [demo](http://prayashm.github.io/PapaParse/demo)
#218


mholt:
Hey @prayashm, thanks for the clean PR and test case, and the demo!

I'm not too fond of making it so easy to break the standard (quotes are `"` and that's that), but what's new. 😛

Issue #290, opened 2016-02-18
Papa fails to recognize correct delimiter


egeriis:
In this CSV document, Papa recognizes comma `,` as the delimiter, while the correct one is semicolon `;`.

```
08/01/2016;A;-21.419,06
12/01/2016;A;-21.375,00
13/01/2016;A;-15.000,00
14/01/2016;A;-1.071,31
15/01/2016;A;37.996,97
19/01/2016;A;-53,9
26/01/2016;A;-10.301,58
28/01/2016;A;23.750,00
28/01/2016;A;-1.341,99
29/01/2016;A;-7.612,50
29/01/2016;A;2.031,00
04/01/2016;A;-317,94
26/01/2016;A;-308,33
27/01/2016;A;-4.000,00
27/01/2016;A;-4.000,00
27/01/2016;A;-3.138,32
29/01/2016;A;-4.000,00
15/01/2016;A;-116
20/01/2016;A;-2.405,17
04/01/2016;A;-249
04/01/2016;A;1.509,88
07/01/2016;A;-275
18/01/2016;A;-3.172,95
```

I had a brief look at `guessDelimiter`, and `delta` seems to be the issue, due to some rows having only 1 column with comma `,` as delimiter.

Could it possibly be a factor to take into account, how many columns are detected with the different delimiters, the more the better?

mholt:
> Could it possibly be a factor to take into account, how many columns are detected with the different delimiters, the more the better?

Not necessarily, since the data may just contain a lot of delimiters in it.

Frankly I'd recommend not auto-guessing the delimiter if your data has delimiter characters in it; just specify it manually.

Issue #153, opened 2015-01-12
delimiter auto-detect error


imadhajjchahine:
Hi,

I have the following data and trying to auto-detect the delimiter (replaced tab with ; for visibility):

101300;010114;00;2132500450881;A.N. au 010114;;;733480,00;-733480,00
106110;010114;00;2132500450881;A.N. au 010114;;;73348,16;-73348,16
106320;010114;00;2132500450881;A.N. au 010114;;;539346,47;-539346,47

since the numbers are in french notation, the decimal separator used is a comma not a point.

The problem is that papaparse detect the comma as a seperator for the CSV not the tab or semi-column in this example.

I checked the source code for the guessDelimiter function and i think the bestDelimiter should also be determined by the highest avgFieldCount not only the bestDelta.

Waiting for your feedback on the issue.

Thank you.

mholt:
Ah, yep, that's a confusing situation for the parser to figure out by itself. I recommend setting the delimiter manually in those cases.

Thanks for taking a look at the source code! Using highest avgFieldCount may help in your case, but it assumes that the faux-delimiter (the comma in your case) is less frequent than the actual delimiter. Other CSV files may have just two or three actual fields but multiple commas in each field.

PR #318, opened 2016-07-01
Add callback option for specifying delimiter.

mickylad:
I noticed there are some difficulties with auto-detecting delimiters (best described at https://github.com/mholt/PapaParse/issues/153 & https://github.com/mholt/PapaParse/issues/290).

I've faced similar issues, but I can't follow the suggestion of specifying the delimiter every time, because I don't know it before parsing the file.

However I do, in my case, have a good idea of what structure the data will be in - for example I always expect the first cell to be empty (and can handle the errors arising from that ever not being true).

This lead me to the idea of allowing a callback to be used to determine the delimiter, based on the parsed file content (before it is parsed to CSV). This lets me simply return the first character of the input to specify the delimiter, rather than having to hard-code it or rely on auto-detection.

Let me know if you think this is an acceptable solution, and if it can be improved in anyway.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -764,6 +764,11 @@
 				}
 				_results.meta.delimiter = _config.delimiter;
 			}
+			else if(typeof _config.delimiter === 'function')
+			{
+				_config.delimiter = _config.delimiter(input);
+				_results.meta.delimiter = _config.delimiter;
+			}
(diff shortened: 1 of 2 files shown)


pokoli:
Thanks for your contribution!

Issue #246, opened 2015-07-30
PapaParse unparser is not building the header fields dynamically


tejasatara:
Hi,

I have observed that the unparser function takes into consideration only first object in the array of objects to form the header fields. The data I'm using is dynamic and some fields are present in only few objects.

Example: 

```javascript
[{
                        "_index": "documents",
                              "_type": "message",
                              "_id": "4868ecf0-01ba-3f1a-b6f5-e7eb0c9c630f",
                              "_score": null,
                              "sort": [
                                  1438247629232
                              ]
                          },
                          {
                            "_index": "documents",
                            "_type": "url",
                            "_id": "5868ecf0-01ba-3f1a-b6f5-e7eb0c9c630f",
                            "tags": "abc",
                            "_score": null,
                            "sort": [
                               14382476292334341
                               ]
                          }]
```
Here you can see that tags field is only present in the second object and it is being ignored. Is there a way in which the header can be formed dynamically?

mholt:
Yes; you can specify the fields and data separately. From the docs:

> data can be one of: An object explicitly defining fields and data

```
Papa.unparse({
    data: [ ... ],
    fields: [ ... ]
});

**Comment by tejasatara** [comment]
For getting the fields I need to traverse the results and form the array I was hoping if this would happen automatically as a part of the parser.

**Comment by mholt** [comment]
Sorry, but there's multiple ways to flatten data structures, so I'm choosing to leave it up to the user. Thanks for the feedback though, and for using Papa Parse!

**Comment by eschwartz** [comment]
This was a sneaky bug for me to find.

Any chance we could add a note about this to the docs?

```js
// With implicit header row
var csv = Papa.unparse([
	// By default, PapaParse will use the keys of this first
	// object to populate the header row
	{
		"Column 1": "foo",
		"Column 2": "bar"
	},
	{
		"Column 1": "abc",
		"Column 2": "def"
	}
]);
```

Are the docs in a github repo somewhere? I could send you a pull request, if they are

mholt:
Sure, they're in the gh-pages branch. Make sure to put a space between Papa and Parse.

Maybe just do this instead:

```
// With implicit header row (keys of
// first object populates header row)
```

PR #326, opened 2016-07-21
Document how header row is populated

eschwartz:
when using a collection of data objects

See #246


pokoli:
@eschwartz Thanks for contributing!

eschwartz:
ha, that was a long time ago! Thanks for merging it :)

PR #332, opened 2016-08-17
Leading Zero As String

ataft:
When using dynamic typing, set the "leadingZeroAsString" config option to "true" in order to treat values with leading zeros as strings. Used the following RegEx logic:

var regExp = /^0[0-9].*$/
console.log(regExp.test("0001")); // true
console.log(regExp.test("001")); // true
console.log(regExp.test("01")); // true
console.log(regExp.test("00.12312")); // true
console.log(regExp.test("000.1232")); // true
console.log(regExp.test("1")); // false
console.log(regExp.test("20")); // false
console.log(regExp.test("0.1")); // false
console.log(regExp.test("0.123123")); // false

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -701,6 +701,7 @@
 	{
 		// One goal is to minimize the use of regular expressions...
 		var FLOAT = /^\s*-?(\d*\.?\d+|\d+\.?\d*)(e[-+]?\d+)?\s*$/i;
+		var LEADINGZERO = /^0[0-9].*$/;
 
 		var self = this;
 		var _stepCounter = 0;	// Number of times step was called (number of rows parsed)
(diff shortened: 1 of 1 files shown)


pokoli:
Thanks for you pull request. Could you add some unit test so we can ensure that everything is working as expected and that it won't be broken by future changes?

papaparse.js
```diff
@@ -983,7 +984,8 @@
 		function tryParseFloat(val)
 		{
 			var isNumber = FLOAT.test(val);
-			return isNumber ? parseFloat(val) : val;
+			var leadingZeroAsString = _config.leadingZeroAsString&&LEADINGZERO.test(val);
+			return isNumber&&!leadingZeroAsString ? parseFloat(val) : val;
```
pokoli: Could you please add some spaces between && text so it's easier to read? Thanks

pokoli:
I'm wondering if your requirements can be now achieved with #377

pokoli:
I'm closing this in favour of #377

ataft:
@pokoli this can be achieved with #377 only if you know the datatype of the columns. if you don't, then I don't see any other way.

pokoli:
@ataft Indeed you can parse a preview of the file to determine the datatype of the columns, and the pass it to the parser.

PR #333, opened 2016-08-19
check for objects and save the url

jpat1993:
In the loop, I checked the str, that we are adding in the output csv file, to see if it was an object. If it is an object, then it is the image object from the parse pull. Then it will display the url of the image so the users can reference.


pokoli:
I don't think this is a behaviour we want as part of the standard library. Instead of getting the url when parsing the file you should get all the object properties before and pass only the strings to the unparse method.

Issue #341, opened 2016-09-12
Suggestion Header WIDE, Papa.unparse()


kennyyen:
Currently I am using Papa.unparse() to parse an array of object data where I want to keep all the fields but the first object in the array does not necessary has all the properties fields, other object has. 

So, I modified the setting fields part to check every property of every object within the input array.

I totally understand that performance is a big issue here and a lot of user care about it, so I am suggesting to have a special parser, something like WIDE, to switch on and off this functionality of having the complete fields set when objects in the input array does not have equal property lengths.

(Any better idea, or suggestions ?)

In the JsonToCsv(_input, _config)

``` js
            else if (typeof _input[0] === 'object'){
                if (_config && _config.wide){
                    var tmpObj = [];
                    for(var i = 0; i < _input.length; i++){
                        for (var prop in _input[i]) {
                          if (_input[i].hasOwnProperty(prop)) {
                            if (tmpObj.indexOf(prop) < 0) {
                              tmpObj.push(prop);
                            }
                          }
                        }
                    }
                    return serialize(tmpObj.sort(), _input);
                } else
                    return serialize(objectKeys(_input[0]), _input);
            }
```

And in the unpackConfig()

``` js
            if (typeof _config.wide === 1)
                _wide = _config.wide;
```

kennyyen:
If Possible, I would like to implement/add another Header, something like ownFields, which would be an Array of Objects and objects will contain name and value. So, when this header is defined, the Papa.unparse() will use object.name as the displaying field name, and value for actual property matching purpose. Please let me know if you think this is a legit and useful functionality!! Thanks !

pokoli:
@kennyyen why you don't modify the first object of your data ensuring that it has all the names? Indeed that can be done with a code very similar to the one you posted. As you can manipulate the data before unparsing it I think there is no need to include it on PaPaParse.

Hope it helps.

kennyyen:
Oh @pokoli, it is because I would not know how many columns are there in the first place. Which then I will have to loop through the data once, modify the first item, then pass them into PaPaParse and let it be looped through again.

Thank you for your reply tho !

PR #342, opened 2016-09-12
Suggestion-Header-Wide-PapaUnparse-#341

kennyyen:
*Added Header check, wide, possible value = 1
*If wide = 1, check for all properties of all object within the input array
 to construct the header fields


pokoli:
I'm closing it as explained on https://github.com/mholt/PapaParse/issues/341#issuecomment-271814837

PR #346, opened 2016-09-19
[unparse] header-option.

tijszwinkels:
For unparse: use the header: option to specify whether a header should be printed or not.


pokoli:
LGTM

PR #361, opened 2016-12-15
Add trim option

xinlincao:
Support removing whitespaces from both ends of a string value.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -1013,6 +1013,7 @@
 		var preview = config.preview;
 		var fastMode = config.fastMode;
 		var quoteChar = config.quoteChar || '"';
+		var trim = config.trim;
 
 		// Delimiter must be valid
 		if (typeof delim !== 'string'
@@ -1035,6 +1036,15 @@
 		// We're gonna need these at the Parser scope
 		var cursor = 0;
 		var aborted = false;
+
+		var trimElements = function(elementArray) {
+			var i = 0;
+			for (i = 0; i < elementArray.length; ++i) {
+				if (typeof elementArray[i] === 'string') {
+					elementArray[i] = elementArray[i].trim();
+				}
+			}
+		}
 
 		this.parse = function(input, baseIndex, ignoreLastRow)
 		{
@@ -1073,13 +1083,26 @@
 					if (stepIsFunction)
 					{
 						data = [];
-						pushRow(row.split(delim));
+						var rowElements = row.split(delim);
+						if (trim) {
+							trimElements(rowElements);
+						}
+						pushRow(rowElements);
 						doStep();
-						if (aborted)
+						if (aborted) 
+						{
 							return returnable();
+						}
 					}
 					else
-						pushRow(row.split(delim));
+					{
+						var rowElements = row.split(delim);
+						if (trim) {
+							trimElements(rowElements);
+						}
+						pushRow(rowElements);
+					}
+
 					if (preview && i >= preview)
 					{
 						data = data.slice(0, preview);


mholt:
Why can't you just call .trim() yourself? Papa Parse is a CSV parser, not a formatter.

xinlincao:
1. It is very inconvenient and not easy to remove leading/ending whitespaces myself. For instance, if the header option is true, the parsed data will be something like `[{'______FOO____':  '__aa____', '_________BAR____: '____bbb____'}, {'____FOO   ____':  '____cc____', '____BAR____ ': '____ddd ____'}]` (I use underscores to represent whitespaces here). So users need to another round post-processing to get `[{'FOO':  'aa', 'BAR': 'bbb'}, {'FOO':  'cc', 'BAR': 'ddd'}]`. And such post-processing of renaming values and keys may involve deleting and adding keys of objects, not only by calling trim(). Therefore, I found that introducing the "trim" option will make the life easier.

2. Taking whitespace into consideration does not do harm to a parser. Actually, there are examples of how parsers should process the whitespaces, such as C language parser (where whitespaces are not important) and Python parser (where whitespaces matter).

pokoli:
@xinlincao will trim headers solve the issue for you? You can then call the trim manually on all the values if required.

xinlincao:
@pokoli 

I tried and found that it is very inconvenient and not easy to remove leading/ending whitespaces myself. For instance, if the header option is true, the parsed data will be something like [{'______FOO____': '__aa____', '_________BAR____: '____bbb____'}, {'____FOO ____': '____cc____', '____BAR____ ': '____ddd ____'}] (I use underscores to represent whitespaces here). So users need to another round post-processing to get [{'FOO': 'aa', 'BAR': 'bbb'}, {'FOO': 'cc', 'BAR': 'ddd'}]. And such post-processing of renaming values and keys may involve deleting and adding keys of objects, not only by calling trim(). Therefore, I found that introducing the "trim" option will make the life easier.

Should you have a simple and elegant way to trim headers, please let me know. Thanks!

pokoli:
@xinlincao I'm wondering if you can post a an example file and explain the behaviour with the PR and without it. As it's not clear to me what you are trying to achieve.

xinlincao:
@pokoli 
A simple example is:
Step 1: Go to http://papaparse.com/demo
Step 2: Put the following data in the text box:
<pre>   Column 1       ,Column 2    ,        Column 3,    Column 4
    1-1    ,1-2,    1-3     ,1-4
2-1    ,    2-2,    2-3,2-4
3-1,3-2,3-3,3-4
4,5,6,    7</pre>

Noting that there are leading spaces and ending spaces in the data.
Step 3: Click the "Header row" check box to enable header option.
Step 4: Click "Parse" button.

In the devtool console, you see the result data is something like:
`
Array[4]
0
:
Object
" Column 3"
:
"    1-3     "
" Column 4"
:
"1-4"
" Column 1 "
:
"    1-1    "
"Column 2 "
:
"1-2"
__proto__
:
Object
1
:
Object
" Column 3"
:
"    2-3"
" Column 4"
:
"2-4"
" Column 1 "
:
"2-1    "
"Column 2 "
:
"    2-2"
__proto__
:
Object
2
:
Object
" Column 3"
:
"3-3"
" Column 4"
:
"3-4"
" Column 1 "
:
"3-1"
"Column 2 "
:
"3-2"
__proto__
:
Object
3
:
Object
" Column 3"
:
"6"
" Column 4"
:
"    7"
" Column 1 "
:
"4"
"Column 2 "
:
"5"
__proto__
:
Object
`

Now you can see that the object keys and values contain leading and ending spaces, which is something I want to avoid, especially the keys. 

Thank you!

pokoli:
@xinlincao and why you can not trim the values after parsing? 

For example you have the field names with spaces as: 

> results.meta.fields

This values can be used to trim the headers (and keep a mapping of the current column values). See http://papaparse.com/docs#meta for further reference. About row values they can be trimed when reading them. 

I think there is no reason for adding a trim option. I understand that it can be easier to implement to trim directly when parsing, but this will be out of the scope of the library as @mholt previously explained.

SelaO:
To handle those space you can trim the whole file out of spaces: `.replace(/ /g, '')`

PR #365, opened 2017-01-31
Determine global scope without using eval

### LINKED ISSUE #363: Don't use eval to determine global scope
This commit https://github.com/mholt/PapaParse/commit/fbf22867bfc6280192c5267cab84eadf50307170 use eval to determine the global scope. This will break many things (Content Security Policy is one).

Instead you can use something like this:
https://github.com/paulmillr/es6-shim/commit/2367e0953edd01ae9a5628e1f47cf14b0377a7d6

Would you be willing to make this change?

sandstrom:
This is makes PapaParse useable in sites where Content Security Policy is used.

Closes #363

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -26,7 +26,19 @@
 {
 	'use strict';
 
-	var global = Function('return this')();
+	var global = (function () {
+		// alternative method, similar to `Function('return this')()`
+		// but without using `eval` (which is disabled when
+		// using Content Security Policy).
+
+		if (typeof self !== 'undefined') { return self; }
+		if (typeof window !== 'undefined') { return window; }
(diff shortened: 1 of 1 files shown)


pokoli:
@sandstrom this is breaking test suit, could you fix please?

sandstrom:
@pokoli Sorry, I don't know why this wouldn't work. AFAIK it works in a browser and node. Perhaps there is something peculiar with the test environment?

pokoli:
@sandstrom thanks for you're patch. I have merged them manually and fixed the test with https://github.com/mholt/PapaParse/commit/ff398191f2c0547b0164e487414aec37a030bf65

sandstrom:
@pokoli Thanks, that's awesome! ⛵️

PR #370, opened 2017-03-04
Node support

LeviticusMB:
Adds support for parsing Node streams.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -68,6 +68,7 @@
 	Papa.NetworkStreamer = NetworkStreamer;
 	Papa.FileStreamer = FileStreamer;
 	Papa.StringStreamer = StringStreamer;
+	Papa.ReadableStreamStreamer = ReadableStreamStreamer;
 
 	if (global.jQuery)
 	{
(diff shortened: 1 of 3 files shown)


papaparse.js
```diff
@@ -226,6 +227,10 @@
 			else
 				streamer = new StringStreamer(_config);
 		}
+		else if (_input.readable === true && typeof _input.read === 'function' && typeof _input.on === 'function')
```
pokoli: Is there any special reason to add this before the global.File If?

For me it sounds more logical to add new features at the end.

papaparse.js
```diff
@@ -725,6 +730,76 @@
...
+			}
+			catch (error)
+			{
+				this._streamError(error);
+			}
+		}, this);
+
+		this._streamError = bindFunction(function(error)
+		{
+			this._input.removeListener('data', this._streamData);
```
pokoli: As you are removing the same Listeners on Error and on End, I'm wondering if it makes sense to create a function to remove the listeners and call them. This will ease the maintenance in case of adding new listeners in the future.

papaparse.js
LeviticusMB: Yes, because of `|| _input instanceof Object` on the line you're referring to.

pokoli: Acknowledged.

pokoli:
Hi @LeviticusMB I've squashed all the commits into a single one and merged the Pull Request.

Thanks for contributing. 

I will try to do a major release including this functionality in the next days.

oliviertassinari:
It would be great to document the usage of this feature. I'm doing the following:
```js
function parseCSV(filename) {
  return new Promise((resolve, reject) => {
    Papa.parse(fs.createReadStream(filename), {
      complete(parsedCsv) {
        resolve(parsedCsv)
      },
      error: reject,
    })
  })
}
```

PR #375, opened 2017-03-23
Add support for download xhr headers

### LINKED ISSUE #374: http headers when download is true
I may have missed this from the docs, but if I want to set an Authorization header on the request when streaming a file, is there a way to do so?

eg:
```
Papa.parse( 'some/endpoint.csv', {
	download: true,
	//downloadConfig: {headers:{authorization:'some token'}} was thinking an option like this may exist but haven't found reference to anything
	step: data => {
		console.log( data );
	},
	complete: this._subject.onCompleted
} );
```

j--w:
Fixes #374

As discussed in issue, tests for NetworkStream require additional infrastructure to be addressed in a new issue.

I could not find any documentation about whether the minified file should be generated in this commit or if there is another process for that. Please advise.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -581,7 +581,7 @@
 			}
 
 			xhr.open('GET', this._input, !IS_WORKER);
-
+			// Headers can only be set when once the request state is OPENED
 			if (this._config.downloadRequestHeaders)
 			{
 				var headers = this._config.downloadRequestHeaders;
--- a/papaparse.js
+++ b/papaparse.js
@@ -574,6 +574,14 @@
 				xhr.withCredentials = this._config.withCredentials;
 			}
 
+			if (!IS_WORKER)
+			{
+				xhr.onload = bindFunction(this._chunkLoaded, this);
+				xhr.onerror = bindFunction(this._chunkError, this);
+			}
+
+			xhr.open('GET', this._input, !IS_WORKER);
+
 			if (this._config.downloadRequestHeaders)
 			{
 				var headers = this._config.downloadRequestHeaders;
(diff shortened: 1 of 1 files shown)


papaparse.js
```diff
@@ -582,6 +582,19 @@
 
 			xhr.open('GET', this._input, !IS_WORKER);
 
+			if (this._config.downloadRequestHeaders)
```
pokoli: I find it more logical to set it after withCredentials as we are setting up the request there.

papaparse.js
```diff
@@ -582,6 +582,19 @@
 
 			xhr.open('GET', this._input, !IS_WORKER);
 
+			if (this._config.downloadRequestHeaders) 
+			{
+				var headers = this._config.downloadRequestHeaders;
+
+				for (var headerName in headers) 
+				{
+					if (Object.prototype.hasOwnProperty.call(headers, headerName))
```
pokoli: why is this required? Is not enought by setting the setRequestHeader?

papaparse.js
j--w: The xhr state must be `OPENED` to set headers.

papaparse.js
j--w: Good point, I've removed this redundant check

papaparse.js
```diff
@@ -582,6 +582,16 @@
 
 			xhr.open('GET', this._input, !IS_WORKER);
 
+			if (this._config.downloadRequestHeaders)
```
pokoli: Can we add a comment to indicate that the state must be OPEN to set the request?

pokoli: s/request/header/ sorry :confused:

j--w: No worries, added in latest changes

pokoli:
Code have been merged. Thanks for contributing!

PR #377, opened 2017-04-04
add ability to configure which columns are dynamically typed

vyrak:
Goals are similar to #332.

- supports configuring `dynamicTyping` with a callback, giving more flexibility on what to dynamically type without needing to be strict on the fields you want to configure

See tests for usage/examples.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -426,6 +426,10 @@
 			this._isAllDynamic = true;
 			return;
 		}
+		else if (typeof config === 'function')
+		{
+			this._isIncluded = config;
+		}
 		else if (!config)
 		{
 			return;
@@ -456,6 +460,10 @@
 	DynamicTypingConfig.prototype.constructor = DynamicTypingConfig;
 	DynamicTypingConfig.prototype.isIncluded = function(field)
 	{
+		if (typeof this._isIncluded === 'function')
+		{
+			return this._isIncluded(field);
+		}
 		return this._isAllDynamic || this._columnConfig[field] === true || (this._isBlackListOnly && this._columnConfig[field] !== false)
 	};
 
--- a/papaparse.js
+++ b/papaparse.js
@@ -417,54 +417,20 @@
 
 	function DynamicTypingConfig(config)
 	{
-		this._columnConfig = {};
-		this._isAllDynamic = false;
-		this._isBlackListOnly = true;
-
-		if (config === true)
-		{
-			this._isAllDynamic = true;
-			return;
-		}
-		else if (typeof config === 'function')
-		{
-			this._isIncluded = config;
-		}
-		else if (!config)
-		{
-			return;
-		}
-
-		if (config.__only instanceof Array)
-		{
-			this._isBlackListOnly = false;
-			config.__only.forEach(function(field)
-			{
-				this._columnConfig[field] = true;
-			}.bind(this));
-		}
-		else
-		{
-			this._isBlackListOnly = !!config.__except;
-			this._columnConfig = config;
-		}
-
-		if (config.__except instanceof Array)
-		{
-			config.__except.forEach(function(field)
-			{
-				this._columnConfig[field] = false;
-			}.bind(this));
-		}
+		this._config = config;
 	}
 	DynamicTypingConfig.prototype.constructor = DynamicTypingConfig;
 	DynamicTypingConfig.prototype.isIncluded = function(field)
 	{
-		if (typeof this._isIncluded === 'function')
+		if (typeof this._config === 'function')
 		{
-			return this._isIncluded(field);
+			return this._config(field);
 		}
-		return this._isAllDynamic || this._columnConfig[field] === true || (this._isBlackListOnly && this._columnConfig[field] !== false)
+		if (typeof this._config === 'object')
+		{
+			return this._config[field] === true;
+		}
+		return !!this._config;
 	};
 
 	/** ChunkStreamer is the base prototype for various streamer implementations. */
--- a/papaparse.js
+++ b/papaparse.js
@@ -424,7 +424,7 @@
 	{
 		if (typeof this._config === 'function')
 		{
-			return this._config(field);
+			return this._config(field) === true;
 		}
 		if (typeof this._config === 'object')
 		{
--- a/papaparse.js
+++ b/papaparse.js
@@ -161,7 +161,7 @@
 				queue.splice(0, 1);
 				parseNextFile();
 			}
-		}
+		};
 	}
 
 
@@ -193,7 +193,7 @@
 	function CsvToJson(_input, _config)
 	{
 		_config = _config || {};
-		_config.dynamicTyping = _config.dynamicTyping || false;
+		_config.dynamicTyping = new DynamicTypingConfig(_config.dynamicTyping);
 
 		if (_config.worker && Papa.WORKERS_SUPPORTED)
 		{
@@ -414,6 +414,50 @@
 			return false;
 		}
 	}
+
+	function DynamicTypingConfig(config)
+	{
+		this._columnConfig = {};
+		this._isAllDynamic = false;
+		this._isBlackListOnly = true;
+
+		if (config === true)
+		{
+			this._isAllDynamic = true;
+			return;
+		}
+		else if (!config)
+		{
+			return;
+		}
+
+		if (config.__only instanceof Array)
+		{
+			this._isBlackListOnly = false;
+			config.__only.forEach(function(field)
+			{
+				this._columnConfig[field] = true;
+			}.bind(this));
+		}
+		else
+		{
+			this._isBlackListOnly = !!config.__except;
+			this._columnConfig = config;
+		}
+
+		if (config.__except instanceof Array)
+		{
+			config.__except.forEach(function(field)
+			{
+				this._columnConfig[field] = false;
+			}.bind(this));
+		}
+	}
+	DynamicTypingConfig.prototype.constructor = DynamicTypingConfig;
+	DynamicTypingConfig.prototype.isIncluded = function(field)
+	{
+		return this._isAllDynamic || this._columnConfig[field] === true || (this._isBlackListOnly && this._columnConfig[field] !== false)
+	};
 
 	/** ChunkStreamer is the base prototype for various streamer implementations. */
 	function ChunkStreamer(config)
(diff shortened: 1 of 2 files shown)


papaparse.js
```diff
@@ -161,7 +161,7 @@
 				queue.splice(0, 1);
 				parseNextFile();
 			}
-		}
+		};
```
pokoli: This seems not related to this PR

papaparse.js
```diff
@@ -415,6 +415,50 @@
...
+		}
+		else if (!config)
+		{
+			return;
+		}
+
+		if (config.__only instanceof Array)
+		{
+			this._isBlackListOnly = false;
+			config.__only.forEach(function(field)
```
pokoli: Why prefixing with __ ? I think its better to name the variables without __

vyrak: I agree that it looks better, but I think it's necessary to be backwards with the old way of configuring things. If a file happens to have columns named `'only'` or `'except'`, then it would try to cast those column types.

If we decide that we want to remove those underscores, then we should probably deprecate the old way of configuring `dynamicTyping` with a hash & remove that support in the next major release.

Thoughts?

pokoli: Don't understand why it will create conflicts with old way. If config is true or false it will be handled by the lines above and as a  `return` statement exists, this lines will be never executed. And I don't understand what is the realtion with `__parsed_extra`

papaparse.js
```diff
@@ -415,6 +415,50 @@
...
+			{
+				this._columnConfig[field] = true;
+			}.bind(this));
+		}
+		else
+		{
+			this._isBlackListOnly = !!config.__except;
+			this._columnConfig = config;
```
pokoli: Which is the objective of this line? For me the column config must be always a dict containing the expected only and the except attributes. Nothing else.

pokoli: Indeed, I will remove the full else statement.

papaparse.js
```diff
@@ -415,6 +415,50 @@
 		}
 	}
 
+	function DynamicTypingConfig(config)
+	{
+		this._columnConfig = {};
+		this._isAllDynamic = false;
```
pokoli: isAllDynamic is always false, so we can remove it.

papaparse.js
```diff
@@ -415,6 +415,50 @@
...
+			config.__except.forEach(function(field)
+			{
+				this._columnConfig[field] = false;
+			}.bind(this));
+		}
+	}
+	DynamicTypingConfig.prototype.constructor = DynamicTypingConfig;
+	DynamicTypingConfig.prototype.isIncluded = function(field)
+	{
+		return this._isAllDynamic || this._columnConfig[field] === true || (this._isBlackListOnly && this._columnConfig[field] !== false)
```
pokoli: With latest comments, I think we can also remove isBlackListOnly

papaparse.js
```diff
@@ -193,7 +193,7 @@
 	function CsvToJson(_input, _config)
 	{
 		_config = _config || {};
-		_config.dynamicTyping = _config.dynamicTyping || false;
+		_config.dynamicTyping = new DynamicTypingConfig(_config.dynamicTyping);
```
pokoli: I doubt if we should test if the dinamicTyping has the `isIncluded` function, so the user can build it's own `isIncluded` function to return if some field is included or not, in case only and except is not enough.

vyrak: In this case, `dynamicTyping` will always have `isIncluded`, since it's on the prototype. The original `dynamicTyping` config the user provided is always managed through the `DynamicTypingConfig` prototype.

As far as giving the user the ability to provide their own `isIncluded`, that's an interesting idea I didn't consider. Maybe if the user sets `dynamicTyping` as a function, then the `DynamicTypingConfig` object can use that instead. I'll see about adding that.

papaparse.js
vyrak: Not necessarily, if `dynamicTyping` was set to `true`, then it will be marked as `true`. There are tests for that already. See line 424.

papaparse.js
vyrak: Needed for when only the `'__except'` options is provided. It's for the "dynamically type all columns except certain ones" scenario. Test cases [here](https://github.com/vyrak/PapaParse/blob/98dc2405a72c4598a599f96c946c61758a070e55/tests/test-cases.js#L705) and [here](https://github.com/vyrak/PapaParse/blob/98dc2405a72c4598a599f96c946c61758a070e55/tests/test-cases.js#L732)

papaparse.js
vyrak: Needed to be backwards compatible with the old way of configuring. Again, see  [here](https://github.com/vyrak/PapaParse/blob/98dc2405a72c4598a599f96c946c61758a070e55/tests/test-cases.js#L664) &  [here](https://github.com/vyrak/PapaParse/blob/98dc2405a72c4598a599f96c946c61758a070e55/tests/test-cases.js#L682) for use cases.

papaparse.js
vyrak: See [here](https://github.com/vyrak/PapaParse/blob/98dc2405a72c4598a599f96c946c61758a070e55/tests/test-cases.js#L664) for example of old hash config with specific columns configured. Notice how it's possible that a header can be named `'only'` or `'except'`; not likely, but possible.

If you scroll down [further](https://github.com/vyrak/PapaParse/blob/98dc2405a72c4598a599f96c946c61758a070e55/tests/test-cases.js#L682), you can see an example with the `'__parsed_extra'` working with specific columns configured.

pokoli:
Hi, 

After a full review of the code I think that the cleaner solution is to allow to pass a function as dynamicTyping. This function will be called with the field name (or field number if no header). It must return a boolean that will indicate if the dynamic typing must be applied. 

Then we should forget about the only and the except attributes, as they can be implemented directly with the function: For example:

`function only(name) { return name in [list, of, only, fields] }`

The except function will be the same but negating the operator. 

This will give the library user the full power to decide if the dynamic typing should be applied or not for each field, preserving the current behaviour for the simpler cases. 

P.S: I doubt if it will be useful to pass also the parsed value as second argument to the function to determine the dynamic typing, so the user can also chose to skip dynamic typing for some kind of values. Thoughts?

papaparse.js
```diff
@@ -415,6 +415,24 @@
 		}
 	}
 
+	function DynamicTypingConfig(config)
```
pokoli: Why is this class required? I think we can directly manage the isIncluded function in our parser and this will simplify the code.

tests/test-cases.js
```diff
@@ -694,6 +694,24 @@ var PARSE_TESTS = [
...
+		description: "Dynamic typing by indices can be determined by function",
+		input: '001,002,003',
+		config: { dynamicTyping: function(field) { return (field % 2) === 0; } },
+		expected: {
+			data: [[1, "002", 3]],
+			errors: []
+		}
+	},
+	{
+		description: "Dynamic typing by headers can be determined by function",
```
pokoli: This test is duplicated with the previous one, so I think we can remove it. The behaviour of Paparse will be exactly the same if headers are present or not.

papaparse.js
```diff
@@ -969,7 +987,7 @@
 
 		function parseDynamic(field, value)
 		{
-			if ((_config.dynamicTyping[field] || _config.dynamicTyping) === true)
+			if (_config.dynamicTyping.isIncluded(field))
```
pokoli: I'm wondering if we should create a cache to prevent evaluating the function for each field. This will make parsing faster.

vyrak: The only time I think we need to worry about caching is if the user provides a slow callback; still, that's something they can cache themselves.

papaparse.js
vyrak: It normalizes the different types (functions, objects, or boolean) a user can configure `dynamicTyping` with.

tests/test-cases.js
vyrak: I don't think it's a duplicate. They are two different scenarios that appear similar. On one hand, the callback takes an integer for parsing without headers. On the other, the callback takes a string for parsing with headers. It doesn't hurt to document those two scenarios in the specs.

papaparse.js
pokoli: As one of the objectives of Paparse is to be fast, I think we should cache it to prevent the user complaing about slow callbacks. Maybe it can be documented (see gh-pages branch)

papaparse.js
pokoli: Still no need for a class. Just allow a user function and if no function is provided use the current one with the config object.

tests/test-cases.js
pokoli: The code is the same using integers and strings, so no need to test anything else. For documentation use the gh-pages branch which contains the documentation.

papaparse.js
vyrak: Not sure what you mean....do you mind taking over this PR?

pokoli: I will finish the PR, thanks for your work!

pokoli:
I manually merged it in f8255fd and fixed my comments on 04efec0d61d3e20309ab2c7d379d1de9846d7930

Thanks for contributing

vyrak:
Thanks for taking care of that. I do want to note that the cache will we ignored if the callback returns a falsy value. Also, the formatting was changed (tabs vs spaces).

pokoli:
Hi, you are right about the cache. I've fixed it with b04fc00d0dee2e6ded7c1cd217f2c5815f710073 and released a new patch version which contain the fix. Thanks for your feedback.

schmidtk:
@pokoli The dynamic typing function [added to the config here](https://github.com/mholt/PapaParse/commit/04efec0d61d3e20309ab2c7d379d1de9846d7930#diff-eb20bdfe95a768f4eecdf3bebb522a4eR204) is causing an error to be thrown when using Workers. Objects passed to workers cannot have functions in them, or they cannot be cloned.

pokoli:
@schmidtk It should be fixed with 0353c413e1216beb450f06ef4332c6538e25aacd
I also uploaded a new release on npm which contains the fix. 

Thanks for reporting and sorry for the inconvenience.

schmidtk:
Thanks for the fix. Correct me if I'm wrong, but it looks like you can provide a function for `config. dynamicTyping` [line 197](https://github.com/mholt/PapaParse/blob/0353c413e1216beb450f06ef4332c6538e25aacd/papaparse.js#L197) which would have the same result when using workers.

pokoli:
@schmidtk Yes, you can provide a function for dynamic typing and this will break when using workers. But I think we can not do nothing to prevent the application break with bad configuration. 

Any sugesstion is welcome, but please use a new issue for new suggestions. 

Thanks

PR #380, opened 2017-04-13
Update README.md

MTechDE:
Updated the npm install babyparse part

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ Papa Parse can parse a [Readable Stream](https://nodejs.org/api/stream.html#stre
 Additionally, [Rich Harris](https://github.com/Rich-Harris) forked this project to make **[Baby Parse](https://github.com/Rich-Harris/BabyParse)** which runs in Node.js environments.
 
 ```bash
-$ npm install babyparse
+$ npm install babyparse --save
 ```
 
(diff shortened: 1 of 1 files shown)


pokoli:
Hi, 

Thanks for the PR, could you elaborate why the --save option is required?

MTechDE:
If you don't use the --save flag, the package will not be shown in the package.json file.

pokoli:
It's not mandatory to have a package.json file for installing babyparse, so I don't think we must include the --save flag. 

Thanks for the proposal, but I'm closing it as it's not requried and may confuse users that does not have package.json file.

MTechDE:
If you install a package by npm you always have the package.json file.
For example, you install it without --save and share it with your Git and another person wants to work with this project by running npm install, the package is missing and he has to installed it again manually.
So npm packages should be installed with the --save flag.

pokoli:
@MTechDE This is out of the scope of the PapaParse library. managing package.json files is something related to npm, so people using it should read its documentation. 

Also people can install babyparse to parse a csv from a shell script using node and not pack this script in a npm package as they do not want to share with others. So I do not think we must enforce this behaviour to users. 

It's a good practice by people having npm packages, but we should not enforce people to use npm packages.

PR #392, opened 2017-06-06
Better recognition of node commonJS

### LINKED ISSUE #391: Papa not working in electron
Hello,

Papa isn't recognized in electron with the plain error
```
Uncaught ReferenceError: Papa is not defined
    at <anonymous>:1:1
```

While I'm not able to debug further (due to time and the fact that I don't know the codebase)
 but here's an example electron app to demonstrate [PapaTest.zip](https://github.com/mholt/PapaParse/files/1055539/PapaTest.zip)
(`unzip PapaTest.zip && cd PapaTest && npm i && electron .`)

svbeon:
resolves #391, tested with chrome (59.0.3051.3), safari (10.0.2), electron (1.6.10), Legacy node (0.12.14), LTS node (6.10.3) and current node (8.0.0)

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -11,7 +11,7 @@
 		// AMD. Register as an anonymous module.
 		define([], factory);
 	}
-	else if (typeof module === 'object' && module.exports)
+	else if (typeof module === 'object' && typeof exports !== 'undefined')
 	{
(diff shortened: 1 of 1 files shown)


pokoli:
I tested on node (v.4) and firefox and it works as expected. Thank you for the patch.

PR #393, opened 2017-06-14
Support for user-configuration of multiple potential delimiters

stu-blair:
* Currently the 'delimiter' config option lets the user either specify
  the exact specific expected delimiter or pass in a function that
  entirely overrides PapaParse's delimiter-guessing function. I have a
  use-case where I want to use the built-in guessing function, but
  with a restricted set of potential delimiters rather than the default
  set. I imagine other users may also want to restrict this set, or
  otherwise customize it to only consider certain delimiters (such as
  unusual delimiters not currently in the default set).
* This change enables a third option for the 'delimiter' config setting,
  other than a string or function-- if the user passes in an array of
  strings, then that array will be used as the set of potential
  delimiters for the built-in guessing function.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -876,9 +876,9 @@
 				_config.newline = guessLineEndings(input);
 
 			_delimiterError = false;
-			if (!_config.delimiter)
+			if (!_config.delimiter || isArray(_config.delimiter))
 			{
-				var delimGuess = guessDelimiter(input, _config.newline);
+				var delimGuess = guessDelimiter(input, _config.newline, _config.delimiter);
 				if (delimGuess.successful)
 					_config.delimiter = delimGuess.bestDelimiter;
 				else
@@ -1040,10 +1040,14 @@
 			return _results;
 		}
 
-		function guessDelimiter(input, newline)
+		function guessDelimiter(input, newline, customDelimChoices)
 		{
-			var delimChoices = [',', '\t', '|', ';', Papa.RECORD_SEP, Papa.UNIT_SEP];
-			var bestDelim, bestDelta, fieldCountPrevRow;
+			var bestDelim, bestDelta, delimChoices, fieldCountPrevRow;
+
+			if (!customDelimChoices)
+				delimChoices = [',', '\t', '|', ';', Papa.RECORD_SEP, Papa.UNIT_SEP];
+			else
+				delimChoices = customDelimChoices;
 
 			for (var i = 0; i < delimChoices.length; i++)
 			{
@@ -1564,6 +1568,11 @@
 	function isFunction(func)
 	{
 		return typeof func === 'function';
+	}
+
+	function isArray(arr)
+	{
+		return Object.prototype.toString.call(arr) === '[object Array]';
 	}
 
 	return Papa;
--- a/tests/test-cases.js
+++ b/tests/test-cases.js
@@ -635,6 +635,24 @@ var PARSE_TESTS = [
 		}
 	},
 	{
+		description: "Custom array of potential delimiters",
+		input: "a\t,b\t,c\t",
+		config: { delimiter: ['|', ';', ','] },
+		expected: {
+			data: [["a\t", "b\t", "c\t"]],
+			errors: []
+		}
+	},
+	{
+		description: "Custom delimiter array, correctly guesses most prevalent one",
+		input: 'a;%b,%c,%d;',
+		config: { delimiter: ['%', ',', ';'] },
+		expected: {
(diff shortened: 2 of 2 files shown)


mholt:
What's the performance hit?

stu-blair:
Are you asking what's the performance hit to PapaParse caused by my change? or are you asking what's the performance hit in my use-case that motivated me to make this change?

pokoli:
I think @mholt is asking about the performance hit caused by your change

pokoli:
@stu-blair do you have more separator that are not included on the default separator list? Maybe we can add them on the base code...

papaparse.js
```diff
@@ -1040,10 +1040,14 @@
...
+		function guessDelimiter(input, newline, customDelimChoices)
 		{
-			var delimChoices = [',', '\t', '|', ';', Papa.RECORD_SEP, Papa.UNIT_SEP];
-			var bestDelim, bestDelta, fieldCountPrevRow;
+			var bestDelim, bestDelta, delimChoices, fieldCountPrevRow;
+
+			if (!customDelimChoices)
+				delimChoices = [',', '\t', '|', ';', Papa.RECORD_SEP, Papa.UNIT_SEP];
+			else
+				delimChoices = customDelimChoices;
```
caverna: About performance hit, looking at this line and above, it's safe to say that won't have any!
Unless the custom array would be a very large array and the file delimiter is one of the last from this array.

stu-blair:
@mholt -- I don't think performance is a concern here.
1. If the user doesn't use the new config option, there will be virtually no affect on performance.
2. If the user passes in a reasonable-size array of potential delimiters (like ~10 or less), the performance should still be approximately the same, as @pokoli points out.
3. If the user passes in a giant array (like 100s of delimiters), that may slow things down, but that would be a very unusual use case (I can't imagine a realistic one) and the user would surely be aware they were trading performance for some special functionality.

@pokoli -- my use case was not that I wanted to support special new separators, but that I wanted the delimiter-guessing function to only consider a restricted set of delimiters. I am working on an app that only supports comma, tab, and semicolon, so I don't want it to parse based on some other separator like `|` just because there were a lot of `|` characters present in the input.
So, potential uses for my proposed change could be
* Restricting the set of delimiters to a subset of the default `[',', '\t', '|', ';', Papa.RECORD_SEP, Papa.UNIT_SEP]`, as I needed to do.
or
* Expanding or altering the set of guessable delimiters to include characters other than the default set of `[',', '\t', '|', ';', Papa.RECORD_SEP, Papa.UNIT_SEP]`.
It just adds a bit more flexibility.

stu-blair:
I am curious about the `Papa.RECORD_SEP` and `Papa.UNIT_SEP` characters. What exactly are those and why are they included in the default guessable delimiter list?

pokoli:
@stu-blair I'm not sure if we should merge the current code. The delimiter as Function functionality was added to support all use cases given the user the full flexibility. So I will prefer to use the Function delimiter instead of adding new code to manage especial cases.  I prefer to stay  to generic instead of having code on the base library to support custom use cases.

Having said that, I'm wondering if the following will work for you: 

- Use a custom function that tries to guess the delimiter (based on the current guessDelimiter code) for the list of delimiters you want. 

The guess delimiter code is not so complex.

What do you think?

stu-blair:
@pokoli , I definitely agree it's best to not add code to support special cases. However, I would say that my use case is not that special. I think I've made it sound more complicated than it is-- let me try to rephrase:
* PapaParse currently supports letting the user say "use this single delimiter" by passing in this config: 
`{delimiter: ","}`
* The proposed change adds support for letting the user say "use any of these delimiters" by passing in this config: 
`{delimiter: [",", ";"] }`

This seems pretty straightforward and not that special of a use case to me.

pokoli:
I don't like adding the option to specify an array it will be easy to create some malicious code, that creates a big array and makes the guessing process last for a lot of time, hanging the user browser. 

That's why I pointed to use the function option. 

I will be happy to expose the guessDelimiter function if you want to reuse it. 

Just to clarify: Your use case is totally understandable, but I fear that we may have some unexpected problems with this feature on the future. As far as there is another way to implement it with the current code, I prefer to not introduce new code. 

I'm closing this PR. Thanks for proposing it and for your time.

stu-blair:
OK, thanks for the consideration.

I have to say though, I am confused by your 'malicious code' concern. The function option has the exact same capabilities of crashing the user's browser. A coder with bad intentions would also have many other ways to easily crash a user's browser, as they have full control over what's happening. The 'array' option would not give them any new malicious power.

PR #404, opened 2017-07-13
Skip correctly empty lines while detecting delimiter, fixes #309

### LINKED ISSUE #309: Undetectable Delimiter error for local files ending with a empty line, even if skipEmptyLines is true
By default excel adds a empty line at the end of exported csv files. If you try to parse these with the local file function and auto detect delimiter you'll end up with an "Undetectable Delimiter" error. Even if you enable skipEmptyLines you'll get the error. 

It can be reproduced with the demo on your website and the attached file.
[Mappe1.zip](https://github.com/mholt/PapaParse/files/236626/Mappe1.zip)

caesarsol:
Parsing was broken with newline-terminated short data with `delimiter: auto` and `skipEmptyLines: true`, as described in #309.

Test added in first commit, failed with the following error:

screen shot 2017-07-13 at 12 03 48 pm

Second commit fixes the bug and passes the test.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -878,7 +878,7 @@
 			_delimiterError = false;
 			if (!_config.delimiter)
 			{
-				var delimGuess = guessDelimiter(input, _config.newline);
+				var delimGuess = guessDelimiter(input, _config.newline, _config.skipEmptyLines);
 				if (delimGuess.successful)
 					_config.delimiter = delimGuess.bestDelimiter;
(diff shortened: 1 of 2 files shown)


pokoli:
Thanks. I've squashed two commits in one and merged it.

caesarsol:
Thanks! 
(I usually split the two commits so one can verify if the tests are failing 😬 )

Issue #426, opened 2017-10-11
Allow blob urls for PAPA.SCRIPT_PATH


andre-f-paggi:
Hi,

I have a weird problem with my application server in which it doesn't allow to get URLs of javascript files with search string in it, because of URL Rewrite configuration.

I'm saying that because this library always puts "?papaworker" when working with web workers.

Example of failure: 
mycompany.com/assets/papaparser.min.js?papaworker

So, what I suggest is to allow blob URLs, since they do not mess with the server.

Today, I can set the PAPA.SCRIPT_PATH variable, but the always appended "?papaworker" messes the blob URL and doens't allow the browser to point to that blob.

andre-f-paggi:
Setting the PAPA.SCRIPT_PATH variable:
![screenshot_1](https://user-images.githubusercontent.com/8538938/31457668-cbfa43d4-ae93-11e7-99f1-edea99e34c1d.png)

Error in the browser:
GET blob:http://mycompany.com/0c6fb13f-4fba-4383-9326-6e063849bf0f?papaworker 404 (Not Found)

pokoli:
Is there any alternative to remove the ?papaworker argument? I will be happy to accept a PR that adds an aditional way of calling workers.

pokoli:
Fixed by #599

PR #428, opened 2017-10-16
Added blob url support

andre-f-paggi:
It allows the usage of blob urls as a way to avoid passing '?papaworker' in the querystring and work with difficult webserver routes.

Issue: #426

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -42,7 +42,7 @@
 
 
 	var IS_WORKER = !global.document && !!global.postMessage,
-		IS_PAPA_WORKER = IS_WORKER && /(\?|&)papaworker(=|&|$)/.test(global.location.search),
+		IS_PAPA_WORKER = IS_WORKER && (/(\?|&)papaworker(=|&|$)/.test(global.location.search) || isBlobURL(global.location)),
 		LOADED_SYNC = false, AUTO_SCRIPT_PATH;
 	var workers = {}, workerIdCounter = 0;
 
@@ -57,6 +57,7 @@
 	Papa.BAD_DELIMITERS = ['\r', '\n', '"', Papa.BYTE_ORDER_MARK];
 	Papa.WORKERS_SUPPORTED = !IS_WORKER && !!global.Worker;
 	Papa.SCRIPT_PATH = null;	// Must be set by your code if you use workers and this lib is loaded asynchronously
+	Papa.SCRIPT_PATH_AS_BLOB_URL = false;	// Should be used when you want to use blob urls as 'SCRIPT_PATH' location.
 
 	// Configurable chunk sizes for local and remote files, respectively
 	Papa.LocalChunkSize = 1024 * 1024 * 10;	// 10 MB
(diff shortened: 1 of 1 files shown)


package-lock.json
```diff
@@ -0,0 +1,2107 @@
+{
```
pokoli: This file should not be included in the PR

papaparse.js
```diff
@@ -57,6 +57,7 @@
 	Papa.BAD_DELIMITERS = ['\r', '\n', '"', Papa.BYTE_ORDER_MARK];
 	Papa.WORKERS_SUPPORTED = !IS_WORKER && !!global.Worker;
 	Papa.SCRIPT_PATH = null;	// Must be set by your code if you use workers and this lib is loaded asynchronously
+	Papa.SCRIPT_PATH_AS_BLOB_URL = false;	// Should be used when you want to use blob urls as 'SCRIPT_PATH' location.
```
pokoli: I think we should always use one method for loading workers. So for me this variable is useless.

papaparse.js
```diff
@@ -42,7 +42,7 @@
 
 
 	var IS_WORKER = !global.document && !!global.postMessage,
-		IS_PAPA_WORKER = IS_WORKER && /(\?|&)papaworker(=|&|$)/.test(global.location.search),
+		IS_PAPA_WORKER = IS_WORKER && (/(\?|&)papaworker(=|&|$)/.test(global.location.search) || isBlobURL(global.location)),
```
pokoli: Indeed insted of testing if it's blob, we should probably use a diferent url. For example ending with "worker"

papaparse.js
```diff
@@ -1468,15 +1469,46 @@
...
+		}
+
 		var w = new global.Worker(workerUrl);
 		w.onmessage = mainThreadReceivedMessage;
 		w.id = workerIdCounter++;
 		workers[w.id] = w;
 		return w;
 	}
 
+	function createDummyBlobUrlFromUrl(script_path) {
```
pokoli: This function should remove the suffix added to know that's a worker and then load the correct javascript file.

pokoli:
Thanks for the PR, I've added some remarks, could you review them please?

papaparse.js
andre-f-paggi: I agree it would be better with just one method, but the blob url support nowadays is not very high.
http://caniuse.com/#feat=bloburls

papaparse.js
andre-f-paggi: Humm, you mean the blob url should end with "worker"?
I don't know if it is possible nowadays.

andre-f-paggi:
Thanks for reviewing and helping me out on one of my firsts PRs. :slightly_smiling_face: 
I answered some remarks you made.

papaparse.js
pokoli: I mean that we should not decide if it's a worked bassed blob, but using a diferent url. 

I think it's legitime to load paparse using a blob. With the current code it will asume allways it's a worker, which may not be true. That's why I'm sugesting to add a sufix to determine if it's worker or not.

papaparse.js
pokoli: I see it's supported on all major browsers. So it's ok for my side.

papaparse.js
andre-f-paggi: You are right, but I don't have any other idea since the suffix didn't work for me.

pokoli: Could you explain what you have tested, what is working and what is not working? 
I can provide a better advice if you give more details :)

andre-f-paggi: Of course,

I tried adding information to the querystring of the blob url.

Ex:
blob:https://example.org/9115d58c-bcda-ff47-86e5-083e9a215304?papaworker=true

But the url goes 404 when I do that, it only accepts the url as is.

andre-f-paggi: Well searching a little more, I can pass other information in the url, but the method is different, is via fragments: 

http://jsfiddle.net/thpf584n/1/

papaparse.js
pokoli: Forget about this comment.

papaparse.js
pokoli: Ok, I see the problem now. Lets forget about the suffix and use diretcly the Blob URL without SCRIPT_PATH_AS_BLOB_URL. If there are some problems with this design we should probably reintroduce the flag to enable blob loading in the future. 

Sorry for not giving the best advices.

pokoli:
Could you please also remove the package-lock.json file?

andre-f-paggi:
Hi, sorry for letting this PR open for so long, I completely forgot about it.

It is ok now?

papaparse.js
pokoli: Can we remove the SCRIPT_PATH_AS_BLOB_URL please?

andre-f-paggi: Man, I don't know. If we remove this, it will always load as a blob url?

Honestly, I kind of lost the interest in making this PR, because it seems like I was the only one having this problem, it was very specific to my situation.

But if you still think it is still useful, I can try removing that.

pokoli: Yes, we should load always as blob url. 

Do you think there is any drawback of doing so?

pokoli:
Replaced by #599

PR #429, opened 2017-10-31
📝  Add instructions to install for ember

TristanToye:
Added some instructions to install this package and use it in an EmberCLI app.

I came across other unmaintained shims to do this, but they seemed unnecessary. I think updating the docs here will help people who are looking for framework-specific solutions.

If there is a better place for these docs please let me know and I'll move them there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,9 +40,49 @@ Get Started
 
 Use [papaparse.min.js](https://github.com/mholt/PapaParse/blob/master/papaparse.min.js) for production.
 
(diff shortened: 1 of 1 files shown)


pokoli:
I don't think we should include it on the README, as the number of install instructions will be too large to include there. 

Could you add it to the docs page on the gh-pages branch?

pokoli:
I'm closing this PR, feel free to create another for the gh-pages branch. 

Thanks.

Issue #430, opened 2017-11-07
Issues parsing a CSV with " and ' as part of the data adjacent to the delimiter ,


lunatixcoder:
When using PapaParse if there is a " after a , delimiter, the remaining file gets loaded into the one value. I thought I fixed it by specifying quoteChar: "'", in the config, but there are fields with values that start with ' right after the delimiter as well.

Examples of lines of data from CSV...

060003T,RACEAIR DESIGNS INC ,'LIL BITTS ,4,7 ,1,2,01,001,CLASS 1,0000,

05618AR,MILLER ,"A" ,4,1 ,1,1,01,001,CLASS 1,0085,

05618AT,TRAVEL AIR ,"R" ,4,1 ,1,1,01,001,CLASS 1,0085,

I do not control the data, so I am forced to use the file AS IS.

If I use quoteChar: '"', I can get around the "A" row and the "R" row, but not the 'LIL BITTS row. How do I fix this?

Thanks!

pokoli:
@lunatixcoder have you tested using a quoteChar diferent than " or '. Probably using | will do the trick.

lunatixcoder:
Ok, that works.  However, that will only work until I get data (which I dont control) that has a | as part of a data field in the first character.  Now the likelihood of that is probably none. BUT...  If I set...

quoteChar: '', // setting quoteChar to empty 

The positive assertion should be not to look for any quote characters in the data. (IMHO)

The issue that occurs is papaparse.js automatically defaults to quoteChar: '"' at line 1155 of the latest commit at least.

1155    var quoteChar = config.quoteChar || '"';

If I change that line to...

1155    var quoteChar = config.quoteChar || '';

papaparse does what I would expect it to.  Is there a reason if you don't specify quoteChar or set it to nothing that it assumes a " as quoteChar?  I know the RFC is very grey on this, but when its ambiguous you should always assume positive assertion by the programmer.

I was able to debug this just now, and I do accept that a change like that may not be the correct way to fix it, but it does seem pretty glaring that if I specify NO quote char it should ignore quote searching within a field and assume everything between the delimiters is just a literal.  

I would think that best practices would say that you should have the ability to set a config of ignoreQuotes: true(false)   or  be able to set a positive assertion of quoteChar of null and in either case it should ignore quotes between delimiters.

pokoli:
I will be happy to accept a PR the allows false as quoteChar and in this case no quote character is used. 

Probably that will fix your use case.

[...]

PR #431, opened 2017-11-07
fix for issue #430 where " or ' exist as valid data adjacent right to…

lunatixcoder:
In some cases data contains leading quotes formed by " or ' and are right adjacent to the delimiter. 

Example:

060003T,RACEAIR DESIGNS INC ,'LIL BITTS ,4,7 ,1,2,01,001,CLASS 1,0000,

05618AR,MILLER ,"A" ,4,1 ,1,1,01,001,CLASS 1,0085,

05618AT,TRAVEL AIR ,"R" ,4,1 ,1,1,01,001,CLASS 1,0085,

In these cases we want to treat all data between commas as data including those characters.  So we need a way to tell papaparse, with positive assertion, to ignore quotes.  However papaparse takes all false statements and defaults the quoteChar field to " which creates issues parsing the data when the above data example is present.

With this change, if quoteChar is not specified in the papaparse config, it evaluates to undefined which is distinctly checkable so we can keep behavior as it.  This then allows you to not specify quoteConfig and it remains at defaulting to " otherwise if you positively assert quoteChar: "" then it tells papaparse to ignore quotes by saying there are none.


papaparse.js
```diff
@@ -269,7 +269,25 @@
 		var _newline = '\r\n';
 
 		/** quote character */
-		var _quoteChar = '"';
+// lunatixcoer: Changed logic so you can ignore quote characters altogether.
```
pokoli: I will prefer to have a simpler comment. Like: "Use default value only if it's not set".

papaparse.js
```diff
@@ -269,7 +269,25 @@
...
+//				we can check for undefined and default it to """, 
+//              otherwise we should take what literally it is set to or if set to "" ignore quotes altogether.
+//
+// Lunatixcoder: removed...
+//
+//      var quoteChar = config.quoteChar || '"';
+//
+//      and replaced with 	
+
+		if ( config.quoteChar == undefined ) {
```
pokoli: Better to use ===

papaparse.js
```diff
@@ -269,7 +269,25 @@
...
+			var quoteChar = '"'; // sets default value to " so it should not effect existing code used by other programs
```
pokoli: There is no need to comment every line.

papaparse.js
```diff
@@ -269,7 +269,25 @@
...
+		} else {
+			var quoteChar = config.quoteChar; // that way if you set quoteChar in your config to a positive assertion of quoteChar: "", it will ignore quotes between delimiters
```
pokoli: Idem about comments

pokoli:
I can not merge it untill tests are fixed. 

BTW, it will be great to have a unit test for this case.

lunatixcoder:
Ok... I think I made the changes you requested.  Im new to pull requests so Im hoping Im doing this correctly.  Can you see the changes I made per your request to change comments?

papaparse.js
pokoli: This comment is still missing

pokoli:
THere is one change missing. 

Also have a look at tests that are failing:

https://travis-ci.org/mholt/PapaParse/builds/298513060


You can run test localy with: 

`npm test`

Just make sure you have node installed.

papaparse.js
```diff
@@ -269,7 +269,13 @@
 		var _newline = '\r\n';
 
 		/** quote character */
-		var _quoteChar = '"';
+
```
pokoli: Empty line can be removed.

lunatixcoder:
Ok... so doing NPM test on my machine worked.  one thing I did was downgrade from 9.0 to 8.9 of Node because NPM said it wasnt compatible with 9.0. Did I do something wrong?

package.json
```diff
@@ -56,5 +56,8 @@
 		"test-phantomjs": "node tests/test.js --phantomjs",
 		"test-node": "mocha tests/node-tests.js tests/test-cases.js",
 		"test": "npm run test-node && npm run test-phantomjs"
+	},
```
pokoli: This should not be included here.

lunatixcoder:
I think the issue was that I DID remove those and it caused it to fail because I used 8.9.0  I reinstalled 9.0.0 and re committed and I think that will pass

lunatixcoder:
Ughhh it didnt. let me try what you suggested

lunatixcoder:
YAY!

pokoli:
Great tests are green. Could you please remove the non relevant files and include only the changes to papaparse.js file?

Then we will be done.

lunatixcoder:
Im not familiar enough with GitHub to understand what you wanted me to do I think.  Or was this what you wanted? Again, this is my first time doing this.

pokoli:
Ok, don't worry. Just be pattient. 

I don't mean you have to remove all the files from the project. But include on the PR only the changes relevant to the PR. 

Don't worry, I'm going to merge it manually. 

Thanks for your time.

lunatixcoder:
Thanks for being so kind and patient with me. I really want to learn this and I know you are not here to teach me.  I think I know what I did wrong if you want me to try again.

pokoli:
I merged it manually on d9c78a04ba046ff097bd77730c9af7ac8686a980

Thanks for your contribution. 

Indeed, I will accept a pull request to explain the change on the docs which are on the gh-pages branch. So you can retry the process :) But just to make sure to include only the minimum changes required to document the behaviour.

lunatixcoder:
I cant figure out how to create a PR with GitHub on the documentation branch.  I cloned it and went to open a pull request and it said I did not have access to publish to that repository.

David

pokoli:
You have to push to your fork and then createa pull request to this repository.

lunatixcoder:
That is the problem.  It wont let me push to my fork.

This is a screen snapshot of what I get when I click on Push origin.

PR #445, opened 2018-01-17
Make skipEmptyLines skip rows full of empty values

shamess:
Currently, skipEmptyLines only skips lines with no content but for the
new line character. Lines with nothing but the delimiters and the new
line character are just as useless, so skip these too make sense.


shamess:
Is this too much of a change for the `skipEmptyLines` option? Would a new `skipEmptyValues` be better?

shamess:
Using `Object.values` actually rather slow! I'll rethink this!

PR #446, opened 2018-01-18
Make skipEmptyLines skip rows full of empty values

shamess:
Currently, skipEmptyLines only skips lines with no content but for the
new line character. Lines with nothing but the delimiters and the new
line character are just as useless, so skipping these too makes sense.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -946,10 +946,13 @@
 				_delimiterError = false;
 			}
 
-			if (_config.skipEmptyLines)
+			if (_config.skipEmptyLines || _config.skipNullLines)
 			{
 				for (var i = 0; i < _results.data.length; i++)
-					if (isEmptyLine(_results.data[i]))
+					if (
+						(_config.skipEmptyLines && isEmptyLine(_results.data[i])) ||
+						(_config.skipNullLines && isNullLine(_results.data[i]))
+					)
 						_results.data.splice(i--, 1);
 			}
 
@@ -1059,11 +1062,14 @@
 
 				for (var j = 0; j < preview.data.length; j++)
 				{
-					if (_config.skipEmptyLines && isEmptyLine(preview.data[j]))
-					{
+					if (
+						(_config.skipEmptyLines && isEmptyLine(preview.data[j])) ||
+						(_config.skipNullLines && isNullLine(preview.data[j]))
+					) {
 						emptyLinesCount++;
 						continue;
 					}
+
 					var fieldCount = preview.data[j].length;
 					avgFieldCount += fieldCount;
 
@@ -1124,6 +1130,17 @@
 		function isEmptyLine(line)
 		{
 			return line.length === 1 && line[0].length === 0;
+		}
+
+		function isNullLine(line)
+		{
+			for (var prop in line)
+			{
+				if (line.hasOwnProperty(prop) && line[prop].length > 0)
+					return false;
+			}
+
+			return true;
 		}
 
 		function tryParseFloat(val)
--- a/papaparse.js
+++ b/papaparse.js
@@ -878,7 +878,7 @@
 			_delimiterError = false;
 			if (!_config.delimiter)
 			{
-				var delimGuess = guessDelimiter(input, _config.newline, _config.skipEmptyLines);
+				var delimGuess = guessDelimiter(input, _config.newline, _config.skipEmptyLines, _config.skipEmptyFieldsLines);
 				if (delimGuess.successful)
 					_config.delimiter = delimGuess.bestDelimiter;
 				else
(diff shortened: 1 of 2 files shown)


shamess:
This is much faster than [my first attempt](https://github.com/mholt/PapaParse/pull/445), but the question about if this is too big a change for `skipEmptyLines` is still open.

pokoli:
Indeed I think lines with empty values may be usefull for somebody. Let me put an example:

I have a survey questions results exported on csv and somebody responded all questions with an empty response. In that case I want the empty to row in order to count the response, but the empty value. 

I'm wondering why you need to remove the empty values on the parser. Could you explain your use case?

shamess:
Our case is that we have customers who give us CSVs which are often exported from other systems. In the case of one customer they exported their data it added on 120k empty lines, like the kind that's being skipped here. In our UI, they're useless. It was surprising that the `skipEmptyLines` config wasn't skipping this kind.

Your use case makes sense too though.

What do you think about adding another configuration variable, to leave it up to the library user? I struggled to think of a name for what this might be though.

pokoli:
Yes, I think we should add another configuration variable. 

For the name what about skipNullLines ?

What I'm sure is that the new option should be documented. We hold the documentation on the gh-pages branch on the same repo. Did you mind creating another PR to update the docs also?

shamess:
I've updated this with the suggested new configuration option.

I'm excited to hear your thoughts!

shamess:
Whilst writing the documentation, I realised that maybe "skipLinesWithEmptyFields" might explain the feature better than "skipNullLines". What do you think, @pokoli?

pokoli:
Isn't this setting a little bit long? Probably it can be reduced to `SkipEmptyFieldsLines`

README.md
```diff
@@ -55,6 +55,8 @@ Tests
 
 Papa Parse is under test. Download this repository, run `npm install`, then `npm test` to run the tests.
 
+The continuous integration server runs these tests using Node 0.12. You should ensure your tests pass locally using that version of Node.
```
pokoli: Not for this PR. If you wan to add this note please add another PR.

papaparse.js
```diff
@@ -878,7 +878,7 @@
 			_delimiterError = false;
 			if (!_config.delimiter)
 			{
-				var delimGuess = guessDelimiter(input, _config.newline, _config.skipEmptyLines);
+				var delimGuess = guessDelimiter(input, _config.newline);
```
pokoli: Why you removing it? Probably better to pass the config object directly. I prefer to not depend too much on global variables.

tests/test-cases.js
```diff
@@ -1046,6 +1055,36 @@ var PARSE_TESTS = [
...
+	},
+	{
+		description: "Skip null lines and empty lines work well with each other",
+		notes: "",
+		input: 'a,b\n,\n\nc,d',
+		config: { skipNullLines: true, skipEmptyLines: true },
+		expected: {
+			data: [['a', 'b'], ['c', 'd']],
+			errors: []
+		}
```
pokoli: Can we add a test to ensure that empty lines are preserved when skipEmptyLines is false but skipNullLlines is true? 

It will be great to test the skipNullLines combined with header option.

papaparse.js
```diff
@@ -946,10 +946,13 @@
 				_delimiterError = false;
 			}
 
-			if (_config.skipEmptyLines)
+			if (_config.skipEmptyLines || _config.skipNullLines)
 			{
 				for (var i = 0; i < _results.data.length; i++)
-					if (_results.data[i].length === 1 && _results.data[i][0] === '')
+					if (
+						(_config.skipEmptyLines && isEmptyLine(_results.data[i])) ||
```
pokoli: Why creating two functions? It won't be simpler to create a function skipLine which makes the computation of both cases? I don't think each case is to complex to have it's own function.

papaparse.js
shamess: Sure thing - I can add it back.

I removed it because I didn't think of _config as a global, but more of a property of the ParseHandle object.

papaparse.js
shamess: I'll change this. I was following some advice that functions are free, so make a bunch of small, dumb ones. But I can see your point that they can sometimes get _too_ dumb.

Are you thinking the signature would look like this?

`skipLines(_results.data[i], _config.skipEmptyLines, _config.skipNullLines)`?

tests/test-cases.js
shamess: Ah. That test was vital! `skipNullLines` is frustratingly also doing the job of `skipEmptyLines`. I'm figuring out a way to differentiate them now.

shamess: Unfortunately, this might not be easy to fix. By the time the data gets to `isEmptyLine`, for an emptyLine and a emptyFieldsLine the data structure looks the same.

For now, my company may just have to stick with the forked version of this library (or live with a few spreadsheets looking odd). I'll add the failing test I have, in case you've any ideas!

pokoli: Not sure if there is a use case for that, just wanted to see what's the behaviour of both flags together. Probably we can keep this behaviour but we should add a warning on the docs that explains that skipNullLines implies also skipEmptyLines

pokoli:
Test suite is failing, so please make it pass otherwise we can not merge this changes.

MonkeyDZeke:
This behavior is covered by the 'greedy' value passed to `skipEmptyLines` now: https://github.com/mholt/PapaParse/pull/544

PR #448, opened 2018-01-25
Note about testing with the right Node version

shamess:
I tested on a much later version of Node in a previous PR, which lead to
an embarassing failing test on CI. ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,8 @@ Tests
 
 Papa Parse is under test. Download this repository, run `npm install`, then `npm test` to run the tests.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -55,6 +55,8 @@ Tests
 
 Papa Parse is under test. Download this repository, run `npm install`, then `npm test` to run the tests.
 
+The continuous integration server runs these tests using Node 0.12. You should ensure your tests pass locally using that version of Node.
+
```
pokoli: Is the empty line really required?

PR #453, opened 2018-02-06
Issue 452 - fixed an issue where parser does not parse the file properly

smitshah91:
fixed an issue where parser does not parse the file properly

1. When there is a quote at the end of the line and there is some extra spaces after closing quote and newline
2. When there is a quote before delimiter and there is some extra spaces after closing quote and next delimiter

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -1285,10 +1285,10 @@
 							continue;
 						}
 
-						var spacesBetweenQuoteAndDelimiter = extraSpacesLengthAfterQuote(nextDelim);
+						var spacesBetweenQuoteAndDelimiter = extraSpaces(nextDelim);
 
 						// Closing quote followed by delimiter or 'unnecessary steps + delimiter'
-						if (input[quoteSearch+1] === delim  || spacesBetweenQuoteAndDelimiter > 0)
+						if (input[quoteSearch+1+spacesBetweenQuoteAndDelimiter] === delim)
 						{
 							row.push(input.substring(cursor, quoteSearch).replace(quoteCharRegex, quoteChar));
 							cursor = quoteSearch + 1 + spacesBetweenQuoteAndDelimiter + delimLen;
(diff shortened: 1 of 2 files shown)


papaparse.js
```diff
@@ -1285,21 +1285,25 @@
 							continue;
 						}
 
-						// Closing quote followed by delimiter
-						if (input[quoteSearch+1] === delim)
+						var spacesBetweenQuoteAndDelimiter = extraSpacesLengthAfterQuote(nextDelim);
+
+						// Closing quote followed by delimiter or 'unnecessary steps + delimiter'
+						if (input[quoteSearch+1] === delim  || spacesBetweenQuoteAndDelimiter > 0)
```
pokoli: Why not summing the spaces to the QuoteSearch?

papaparse.js
```diff
@@ -1285,21 +1285,25 @@
 							continue;
 						}
 
-						// Closing quote followed by delimiter
-						if (input[quoteSearch+1] === delim)
+						var spacesBetweenQuoteAndDelimiter = extraSpacesLengthAfterQuote(nextDelim);
```
pokoli: Probably we can find a shorter name. Why not extraSpaces?

papaparse.js
```diff
@@ -1285,21 +1285,25 @@
...
-							cursor = quoteSearch + 1 + delimLen;
+							cursor = quoteSearch + 1 + spacesBetweenQuoteAndDelimiter + delimLen;
 							nextDelim = input.indexOf(delim, cursor);
 							nextNewline = input.indexOf(newline, cursor);
 							break;
 						}
 
-						// Closing quote followed by newline
-						if (input.substr(quoteSearch+1, newlineLen) === newline)
+						var spacesBetweenQuoteAndNewLine = extraSpacesLengthAfterQuote(nextNewline);
```
pokoli: idem about name

papaparse.js
```diff
@@ -1285,21 +1285,25 @@
...
+
+						// Closing quote followed by newline or 'unnecessary spaces + newLine'
+						if (input.substr(quoteSearch+1, newlineLen) === newline || spacesBetweenQuoteAndNewLine > 0)
```
pokoli: Idem about sum

papaparse.js
```diff
@@ -1385,6 +1389,21 @@
 				lastCursor = cursor;
 			}
 
+			/**
+             * checks if there are extra spaces after closing quote and newLine without any text
+             * if Yes, returns the number of spaces
+            */
+            function extraSpacesLengthAfterQuote(index) {
+                var spaceLength = 0;
+				if (index !== -1) {
```
pokoli: Which caracter are you using for ident? It not rendered correctly on github, but you should use Tabs to ident code. I will expect this line to be aligned with the one upper.

papaparse.js
smitshah91: Agree. was just trying to be specific.

papaparse.js
smitshah91: Good point.

papaparse.js
smitshah91: Using tabs but IDE might have messed it up. will fix it.

Issue #459, opened 2018-03-04
Manage escaped quotes 


marius-wieschollek:
I am using PapaParse 4.3.7 and i want to import a csv file where the quotes inside a column are escaped with a backslash (`\"`) instead of a double quote (`""`).

The file looks like this:
```csv
"Column 1","Column 2","Column 3","Column 4"
"Text 1","Text in \"Quotes\"","Text 3","Text 4"
```

When i try to import it, the error `"Trailing quote on quoted field is malformed"` is in the result, and the line is converted incorrectly. 

Is there an option to set the escape character for quotes?

pokoli:
For me this should is a missing feature that we probably want to implement. It will be great if you can submit a pull request implementing it.

Probably it's not needed to configure the scape character and always use the blackslash. I don't think anyone is going to use a different charater and if it does we can add support for configuration on the future.

marius-wieschollek:
I added the option `escapeChar` to specify a custom escape character.

marius-wieschollek:
I added the test and made a pull request for the documentation.

Usually the escaped quote character is detected by searching for the next occurrence of the quoteChar and then checking if the next character is the quoteChar and if so skipping both.
When a custom escapeChar is used and the quoteChar is found, i check if the previous character is the escapeChar, not the next one, and if so only the current character is skipped. This should by design already prevent all other occurrences of the escapeChar from being altered.

PR #460, opened 2018-03-05
Add option to configure escape character (#459)

marius-wieschollek:
This pull request adds the optional configuration option `escapeChar` to escape the `quoteChar` in a quoted field. This option allows to parse CSV files where a non-standard escape character is used. (p.E. `\"` instead of `""`) .

If `escapeChar` is provided, it may contain a one character long string.
If `escapeChar` is not provided, the `quoteChar` will be used as fallback.
The `escapeChar` must appear before the `quoteChar`.

Two tests have been added to validate the functionality of this feature.
Solves issue #459.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -1158,6 +1158,10 @@
 		} else {
 			var quoteChar = config.quoteChar;
 		}
+		var escapeChar = quoteChar;
+		if (config.escapeChar !== undefined){
+			escapeChar = config.escapeChar;
+		}
 
 		// Delimiter must be valid
 		if (typeof delim !== 'string'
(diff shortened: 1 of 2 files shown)


pokoli:
Hi, thanks for the pull request and sorry for the delay. I've been bussy and I haven't been able to review until now. 

I'm wondering what will happen if the text contains the escapeChar. For example including a field with : _"\\a"_, should be parsed as _"\a"_

Could you please add a test case that tests this behaviour also? 

It will be great if you create also a PR on the gh-pages branch to update the documentation including this new parameter.

marius-wieschollek:
I added the test and made a pull request for the documentation.

Usually the escaped quote character is detected by searching for the next occurrence of the quoteChar and then checking if the next character is the quoteChar and if so skipping both.
When a custom escapeChar is used and the quoteChar is found, i check if the previous character is the escapeChar, not the next one, and if so only the current character is skipped. This should by design already prevent all other occurrences of the escapeChar from being altered.

pokoli:
I've merged it. Thanks for your contribution!

PR #471, opened 2018-04-06
Add currently supported Node.js versions as test target

phanect:


pokoli:
Thanks for your contribution

PR #477, opened 2018-04-09
Add eslint

gabegorelick:
The goal is to make PapaParse friendlier for new developers.

The eslint config was generated from `eslint --init` with some tweaking of the resulting rules to minimize errors.


papaparse.js
```diff
@@ -297,8 +295,8 @@
 
 				if (!_input.fields)
 					_input.fields =  _input.data[0] instanceof Array
-									? _input.fields
-									: objectKeys(_input.data[0]);
+						? _input.fields
+						: objectKeys(_input.data[0]);
```
gabegorelick: We can relax the rules if we're not concerned with indentation like this. But it was easier to just fix this case.

papaparse.js
```diff
@@ -644,8 +642,8 @@
 		{
 			var contentRange = xhr.getResponseHeader('Content-Range');
 			if (contentRange === null) { // no content range, then finish!
-					return -1;
-					}
+				return -1;
```
gabegorelick: Good example of catching stuff that's completely wrong.

papaparse.js
```diff
@@ -725,11 +723,9 @@
 		config = config || {};
 		ChunkStreamer.call(this, config);
 
-		var string;
 		var remaining;
 		this.stream = function(s)
 		{
-			string = s;
```
gabegorelick: This was never read.

papaparse.js
```diff
@@ -1238,7 +1235,7 @@
 
 			var nextDelim = input.indexOf(delim, cursor);
 			var nextNewline = input.indexOf(newline, cursor);
-			var quoteCharRegex = new RegExp(escapeChar.replace(/[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, '\\$&')+quoteChar, 'g');
+			var quoteCharRegex = new RegExp(escapeChar.replace(/[-[\]/{}()*+?.\\^$|]/g, '\\$&')+quoteChar, 'g');
```
gabegorelick: We can relax https://eslint.org/docs/rules/no-useless-escape, but I think the resulting regex is easier to read.

papaparse.js
```diff
@@ -1401,17 +1398,17 @@
 			/**
              * checks if there are extra spaces after closing quote and given index without any text
              * if Yes, returns the number of spaces
-            */
-            function extraSpaces(index) {
-                var spaceLength = 0;
```
gabegorelick: Mixed tabs and spaces

player/player.js
```diff
@@ -1,6 +1,9 @@
+/* global $ Papa */
+/* eslint no-console:0 */
+
 var stepped = 0, chunks = 0, rows = 0;
 var start, end;
-var parser;
+var parser; // eslint-disable-line no-unused-vars
```
gabegorelick: Not sure if this is used as part of `player.html` so I kept it in.

tests/node-tests.js
```diff
@@ -1,64 +1,61 @@
-(function() {
-	"use strict";
```
gabegorelick: I removed the IIFE wrapper. Didn't seem necessary in a node-only context and it caused eslint to flag the use of sync node APIs at the "top level" (since IIFE is not the top level).

tests/test-cases.js
```diff
@@ -825,7 +825,7 @@ var PARSE_TESTS = [
 		input: 'A,B,C\r\n1,2.2,1e3,5.5\r\n-4,-4.5,-4e-5',
 		config: { header: true, dynamicTyping: { A: true, C: true, __parsed_extra: true } },
 		expected: {
-			data: [{"A": 1, "B": "2.2", "C": 1000, "__parsed_extra": [ 5.5 ]}, {"A": -4, "B": "-4.5", "C": -0.00004}],
+			data: [{"A": 1, "B": "2.2", "C": 1000, "__parsed_extra": [5.5]}, {"A": -4, "B": "-4.5", "C": -0.00004}],
```
gabegorelick: We can relax stylistic stuff like this.

gabegorelick:
Tests are failing since eslint doesn't support Node.js 0.12. Since that reached end of life at the end of 2016, I think we should just drop support for that. I'll open a separate PR for that. Otherwise, I can switch to an older eslint version.

pokoli:
I've merged the drop of 0.12, you should rebase to make the test pass.

gabegorelick:
@pokoli Thanks. Tests are passing now.

package.json
```diff
@@ -47,9 +48,10 @@
 		"serve-static": "^1.7.1"
 	},
 	"scripts": {
+		"lint": "eslint --no-ignore papaparse.js Gruntfile.js .eslintrc.js 'player/**/*.js' 'tests/**/*.js'",
```
pokoli: I think we can remove the player folder from lint and focus only on other files. 

Player is here only for local testing

player/player.js
pokoli: I will prefer to keep player as is and ignore it.

package.json
gabegorelick: Done.

pokoli:
As a general note I will prefer to not relax the style and keep the recomended features

Thanks for your time on this.

I will do a last review tomorrow before merging.

PR #479, opened 2018-04-09
Fix pause/abort/resume in streaming mode

gabegorelick:
This is a reworking of #166. That PR is pretty old, but there seemed to be consensus that while it wasn't perfect, it was perhaps good enough to merge.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -425,6 +425,7 @@
 	{
 		this._handle = null;
 		this._finished = false;
+		this._completed = false;
 		this._input = null;
 		this._baseIndex = 0;
 		this._partialLine = '';
@@ -483,7 +484,7 @@
 			else if (isFunction(this._config.chunk) && !isFakeChunk)
 			{
 				this._config.chunk(results, this._handle);
-				if (this._handle.paused())
+				if (this._handle.paused() || this._handle.aborted())
 					return;
 				results = undefined;
 				this._completeResults = undefined;
@@ -495,8 +496,10 @@
 				this._completeResults.meta = results.meta;
 			}
 
-			if (finishedIncludingPreview && isFunction(this._config.complete) && (!results || !results.meta.aborted))
+			if (!this._completed && finishedIncludingPreview && isFunction(this._config.complete) && (!results || !results.meta.aborted)) {
 				this._config.complete(this._completeResults, this._input);
+				this._completed = true;
+			}
 
 			if (!finishedIncludingPreview && (!results || !results.meta.paused))
 				this._nextChunk();
--- a/papaparse.js
+++ b/papaparse.js
@@ -424,7 +424,6 @@
 	function ChunkStreamer(config)
 	{
 		this._handle = null;
-		this._paused = false;
 		this._finished = false;
 		this._input = null;
 		this._baseIndex = 0;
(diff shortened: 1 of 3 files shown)


papaparse.js
```diff
@@ -426,7 +426,6 @@
 	function ChunkStreamer(config)
 	{
 		this._handle = null;
-		this._paused = false;
```
gabegorelick: Per https://github.com/mholt/PapaParse/pull/166/files#r28797862, this variable is never used.

papaparse.js
```diff
@@ -483,10 +482,10 @@
 					finished: finishedIncludingPreview
 				});
 			}
-			else if (isFunction(this._config.chunk))
+			else if (isFunction(this._config.chunk) && !isFakeChunk)
 			{
 				this._config.chunk(results, this._handle);
-				if (this._paused)
+				if (this._handle.paused() || this._handle.aborted())
```
gabegorelick: `|| this._handle.aborted()` is the one addition to this PR from #166. It's necessary to avoid calling `complete` twice when `abort` is called.

gabegorelick:
There's still an issue with `complete` getting called multiple times, but I've only been able to reproduce it if `chunk` calls `pause` *and* `resume` on the same event loop tick. I'll continue looking into it, but since that should be a rare pattern (why `pause` if you're not doing async work?), I think this should be OK to merge as is.

gabegorelick:
> There's still an issue with complete getting called multiple times

Fixed with my latest commit. I added a `_complete` state to track whether `complete` has been called. It may be useful everywhere we call `complete`, but since I only reproduced it on stream chunk, I only added it there.

papaparse.js
```diff
@@ -426,8 +426,8 @@
 	function ChunkStreamer(config)
 	{
 		this._handle = null;
-		this._paused = false;
 		this._finished = false;
+		this._complete = false;
```
pokoli: If this variable is used to indicate that complete has been already called I prefer to name it `completed`

tests/test-cases.js
```diff
@@ -1564,6 +1564,72 @@ var CUSTOM_TESTS = [
...
+			});
+			setTimeout(function() {
+				handle.resume();
+			}, 500);
+		}
+	},
+	{
+		description: "Chunk functions can abort parsing",
+		expected: [
+			[['A', 'b', 'c'], ['d', 'E', 'f'], ['G', 'h', 'i']]
```
pokoli: Shouldn't be better to expect only a partial result?

Otherwise if the handle does not abort it will return the same results and it will not fail when it should

pokoli:
As this work is based on #166 would you mind keeping the original commit and then adding your changes on top of it? This way we keep the attribution to the original author

papaparse.js
gabegorelick: Done.

tests/test-cases.js
gabegorelick: Done. 

As a follow up, using something like Sinon for spies makes sense to verify `abort` was called.

gabegorelick:
> As this work is based on #166 would you mind keeping the original commit and then adding your changes on top of it? This way we keep the attribution to the original author

Done.

gabegorelick:
@pokoli Thanks for your help. Any idea when a new version of PapaParse will be released?

pokoli:
@gabegorelick I had the idea to prepare one major release before your fixes but I had no time. I will like to include your PR in the next release so it depends on the number of PR you open and when we are able to merge them :smile: 

Jokes aside, I'm a little bit busy for the next days (we have a version release of our main software in 12 days). I hope to be able to push a new PapaParse release after it.

I've added a reminded on my calendar on 24/04/18, so expect to have a release at this dates (or beffore). 

Does it sound good to you?

gabegorelick:
@pokoli Sounds good. Let me know if I can help in any way.

PR #481, opened 2018-04-09
Replace connect with express

gabegorelick:
Not a huge functional difference for how we're using it, but express is better maintained and much more active.


pokoli:
I'm not sure if we should merge it as it does not provide any functional change nor any improvement. 

What I see is that [express](https://libraries.io/npm/express) has more required dependencies than [connect ](https://libraries.io/npm/connect). 

From the same page I see that connect has a more recent release than express. 

Could you elaborate why you say it's better maintained and how this should benefit us?

PR #483, opened 2018-04-09
Upgrade grunt and grunt-contrib-uglify

gabegorelick:


pokoli:
Thanks for your contibution. I've merged it manually in 99c6f4d16ea2f2e06595da05626335e711ff24aa because it was conflicts with other merges.

I've also updated the GruntFile as the preserver comments options has been deprecated.

PR #233, opened 2015-07-15
Use Mocha, PhantomJS and TravisCI for testing

Turbo87:
see https://travis-ci.org/Turbo87/PapaParse/builds/71137250

resolves #155

PR #484, opened 2018-04-09
Upgrade phantomjs

gabegorelick:


package.json
```diff
@@ -41,9 +41,9 @@
 		"grunt": "^0.4.5",
 		"grunt-contrib-uglify": "^0.6.0",
 		"mocha": "^3.5.0",
-		"mocha-phantomjs": "^3.5.4",
+		"mocha-phantomjs": "^4.1.0",
 		"open": "0.0.5",
-		"phantomjs": "1.9.1 - 1.9.7-15",
+		"phantomjs-prebuilt": "2.1.16",
```
gabegorelick: Upstream changed the name to `phantomjs-prebuilt`, but otherwise it should be the same. 

The version pinning [was to accommodate `mocha-phantomjs`](https://github.com/mholt/PapaParse/pull/233#issuecomment-121745865), so it shouldn't be needed anymore.

pokoli: Ok about name change. 

I think we should use ^version to pin to all new compatible versions of prebuild package

gabegorelick: Whoops good catch. Fixed.

PR #486, opened 2018-04-10
Enable more linting rules

gabegorelick:
Hopefully these should be fairly uncontroversial.

Follow on from #477.


papaparse.js
```diff
@@ -615,11 +615,11 @@
 				this._chunkError();
 			else
 				this._start += this._config.chunkSize;
-		}
+		};
 
 		this._chunkLoaded = function()
 		{
-			if (xhr.readyState != 4)
+			if (xhr.readyState !== 4)
```
gabegorelick: If `readyState` were the string `"4"`, then this change would break. But I don't think that's possible.

papaparse.js
```diff
@@ -1173,7 +1174,7 @@
 			comments = false;
 
 		// Newline must be valid: \r, \n, or \r\n
-		if (newline != '\n' && newline != '\r' && newline != '\r\n')
+		if (newline !== '\n' && newline !== '\r' && newline !== '\r\n')
```
gabegorelick: Let me know if these were intentional for performance.

tests/test-cases.js
```diff
@@ -1,6 +1,8 @@
+var chai;
+var Papa;
```
gabegorelick: We should think about switching to let/const in the future.

pokoli:
Did you mind adding some notes about the rules that we are checking on the how to contribute section of the README?

gabegorelick:
> Did you mind adding some notes about the rules that we are checking on the how to contribute section of the README?

I'd be happy to, but not sure exactly what I can add above the docs in https://eslint.org/docs/rules/.

Issue #422, opened 2017-09-27
parser.pause with node filestream is unable to resume


rvazarkar:
Calling parser.pause when using a node stream as an input will cause the parser to fail when calling parser.resume. The parser will fail to continue parsing, and will just go into a loop that returns 0 items repeatedly.

```Papa.parse(fs.createReadStream(filename), {
            header: true,
            skipEmptyLines: true,
            dynamicTyping: true,
            chunk: function(results, parser){
                parser.pause();
                console.log(results.data.length);
                parser.resume();
            }
```

This simple PoC demonstrates the issue

idibidiart:
Hi any progress on this?

LogIN-:
Same issue here, cant use this features and that's a mess! 
Maybe this issue is related to same problem:
https://github.com/mholt/PapaParse/issues/164

PR #490, opened 2018-04-17
Fix async pause/resume with ReadableStreamStreamer

trevorharwell:
I discovered this bug when trying to process and upload 1.3 million csv records to an elasticsearch instance.  The csv reader was parsing data into my queue faster than I could upload the data in my datastore and thus the process memory would crash node.  I implemented a pausing system so that the upload queue wouldn't get overloaded but then I noticed that the parser would call `complete` before the entire csv had been processed.  This fix allows for the pause/resume to work correctly when they are triggered in an async fashion.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -753,6 +753,7 @@
 
 		var queue = [];
 		var parseOnData = true;
+		var streamHasEnded = false;
 
 		this.stream = function(stream)
 		{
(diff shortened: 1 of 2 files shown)


trevorharwell:
I think this issue might relate to https://github.com/mholt/PapaParse/issues/422

pokoli:
Thanks for your contribution!

gabegorelick:
Thanks for fixing my bugs 🤦‍♂️

PR #494, opened 2018-04-26
Add node-style streaming support

trevorharwell:
This may be a controversial PR and as such I fully expect good feedback and discussion.  That being said, I believe that this is a very useful PR that allows CSV processing on extremely large files to be done much easier.   The node stream API is very good at what it does and so I wanted to take advantage of that.

I had a use case (for which I mentioned in my previous PR) to take a CSV file with 1.6 million entries (145.5MB) and upload each entry to an elasticsearch service.  Before this feature, I had to implement a queue system that would pause the parser when overloaded and resume the parser when empty.  It's not complicated code but it can be tricky to get right.  With this feature in place, I could get rid of all that code and replace it with the following:
```
fs.createReadStream(CSV_FILE_PATH)
    .pipe(Papa.parse(Papa.NODE_STREAM_INPUT, {
      header: true,
      skipEmptyLines: true,
    }))
    .pipe(new stream.Writable({
      objectMode: true,
      highWaterMark: MAX_QUEUE_WORKERS,
      write(doc, encoding, callback) {
        uploadDoc(doc, callback);
      }
    }));
```

Please let me know if there is anything you would me to change or explain.

I also realize that this change would require modifying the documentation.  Please let me know how I can assist in doing that.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -72,7 +72,7 @@
 	Papa.FileStreamer = FileStreamer;
 	Papa.StringStreamer = StringStreamer;
 	Papa.ReadableStreamStreamer = ReadableStreamStreamer;
-	Papa.createDuplexStream = createDuplexStream;
+	Papa.DuplexStreamStreamer = DuplexStreamStreamer;
 
 	if (global.jQuery)
 	{
@@ -234,7 +234,8 @@
 		{
 			// create a node Duplex stream for use
 			// with .pipe
-			return createDuplexStream(_config);
+			streamer = new DuplexStreamStreamer(_config);
+			return streamer.getStream();
 		}
 		else if (typeof _input === 'string')
 		{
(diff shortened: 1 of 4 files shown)


trevorharwell:
Thanks for getting back to me.  I totally understand that this library was made for browsers and have myself used BabyParse before PapaParse supported node.  If you determine that this change is too far "out there" from the original purpose of the library my only ask would be that you expose some internals so that I could create this same effect in a 3rd party library.

My main interest in the .`pipe` feature is that its design accounts for backpressure.  Here is a really good description from node describing how `.pipe` solves that problem: https://nodejs.org/en/docs/guides/backpressuring-in-streams/#too-much-data-too-quickly

When I first used PapaParse just by uploading my document from the `step` function my process would crash every time.  The parser would supply records from the csv way faster than my uploader could upload data.  To solve that problem I had to implement a queueing system that would pause the parser when it got overloaded.

This was my first implementation using a queue: https://gist.github.com/trevorharwell/6e40d94112cf0523e12a9ca15a926308
It may seem relatively straightforward but it took me some time to get it all right because I had to think of queuing data and not letting that queue get overloaded.  The node stream api takes care of all that for me.

Here is the implementation using streams: 
https://gist.github.com/trevorharwell/1c66ed066893f4ee97941d3f5b1e9845 

Regardless of how you solve the problem, anyone parsing a large amount of csv data in node is going to have to deal with backpressure.  Either you incorporate an api that forces developers to write code that inherently deals with backpressure or you let developers get tripped up by it on accident (like I did) and then have to come up with their own solution.

An alternate solution to using `.pipe` would be to have a callback in the `step` function and if too many rows get queue up waiting for their callback to be called, then the parser would auto-pause itself until that queue drained.

pokoli:
Thanks for the explanation! Now I fully understand the problem and I think that it will be great to add support for the pipe function in order to manage the backpreasure in NodeJS. I don't think it's needed to create a third party library to manage it.

papaparse.js
```diff
@@ -58,6 +58,7 @@
 	Papa.BAD_DELIMITERS = ['\r', '\n', '"', Papa.BYTE_ORDER_MARK];
 	Papa.WORKERS_SUPPORTED = !IS_WORKER && !!global.Worker;
 	Papa.SCRIPT_PATH = null;	// Must be set by your code if you use workers and this lib is loaded asynchronously
+	Papa.NODE_STREAM_INPUT = {};
```
pokoli: I think an empty object is too generic for this constant. Probably a integer (which is unexpected input) will be a better value.

papaparse.js
```diff
@@ -839,6 +848,107 @@
...
+		};
+
+		this._onCsvComplete = function()
+		{
+			// node will finish the read stream when
+			// null is pushed
+			stream.push(null);
+		};
+
+		config.step = bindFunction(this._onCsvData, this);
```
pokoli: What if the user has specified a step function?

papaparse.js
```diff
@@ -839,6 +848,107 @@
...
+		config.complete = bindFunction(this._onCsvComplete, this);
```
pokoli: Idem about complete function

papaparse.js
trevorharwell: I was thinking that if you are in "node stream" mode it would be okay for `step` and `complete` to be disabled.  It would be easy enough to add them back in though.  What would you like me to do?

papaparse.js
trevorharwell: Would `Papa.NODE_STREAM_INPUT = 1` be okay?  My thought with the object is to have a constant that could not be produced outside of the library so there would be no chance that someone would accidentally invoke the node specific streaming.  If someone from a browser were to invoke node streaming an error would be thrown because there is a dependency on the `stream` module provided by node core libs.

papaparse.js
```diff
@@ -839,6 +848,107 @@
 	ReadableStreamStreamer.prototype.constructor = ReadableStreamStreamer;
 
 
+	function DuplexStreamStreamer(_config) {
+		var Duplex = require('stream').Duplex;
```
trevorharwell: @pokoli I just want to draw attention to the fact that this exists.  This is the first external dependency that this library uses.  I don't think it will be an issue because I have buried inside this constructor that SHOULD only be invoked when ran in a node environment.

pokoli: Mmmm... Is not possble to avoid an external dependency? I've review the node api docs and the pipe functions are part of the core solution since long time ago. 

I will prefer to minimize this kind of dependencies.

papaparse.js
pokoli: Ok but I think it will be great to add a note on the README and explain how which node events are equivalent to step and complete.

papaparse.js
pokoli: The constant is a good idea, I prefer to be explicit in this case instead of allowing null values and then having unexpected results. 

1 seems a good value to me.

papaparse.js
trevorharwell: I totally agree with you.  The only way I know off the top of my mind would be to literally copy the code from node's core library into this library.  Duplex is more than just an interface as it handles a lot of the heavy lifting to make streams possible.
Here is the code for duplex which also depends on `stream.Writable` and `stream.Readable`
https://github.com/nodejs/node/blob/master/lib/_stream_duplex.js

papaparse.js
trevorharwell: okay will do

papaparse.js
pokoli: Ok, I misunderstood your comment. I thought this was an external library that was required to be installed from npm. If it's part of node source code it's ok to depend on it as it's a core library supported by node.

The current design is safe enough to not fail when running outside of node. So it's ok for me. That's what I understood on the first review and that's why I didn't say anything about it ;)

pokoli:
Thanks for your contribution!

albertorestifo:
Can this be released?

I just wasted a lot of time because I saw this functionality documented but I couldn't get it to work, only to find out that it hasn't been released to npm yet.

pokoli:
@albertorestifo I've published a new release (4.5.0) on npm containing this functionality and other improvements. 

Sorry for the inconvenience.

PR #496, opened 2018-05-02
player.js unparse header included in config

villeahy:
Added header to config in unparse in player.js.


pokoli:
Thanks for your contribution!

PR #497, opened 2018-05-02
Throw Error objects instead of strings

gabegorelick:
Unlike strings, Error objects allow you to get stack traces. Also, a lot of libraries will complain if they encounter a thrown non-Error object.

This is in line with standard Javascript best practices. Per https://eslint.org/docs/rules/no-throw-literal:

> It is considered good practice to only throw the Error object itself or an object using the Error object as base objects for user-defined exceptions. The fundamental benefit of Error objects is that they automatically keep track of where they were built and originated.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -178,7 +178,7 @@ module.exports = {
 		"no-tabs": "off",
 		"no-template-curly-in-string": "error",
 		"no-ternary": "off",
-		"no-throw-literal": "off",
+		"no-throw-literal": "error",
 		"no-trailing-spaces": "error",
 		"no-undef-init": "error",
 		"no-undefined": "off",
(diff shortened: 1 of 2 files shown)


pokoli:
Hi @gabegorelick, 

Thanks for your PR. I agree this is an API breaking change. 

About the test passing, maybe it's because errors are stringified on tests. See:

https://github.com/mholt/PapaParse/blob/9155dd5b747cd659a41e0fe760148ee71f89a4e4/tests/test-cases.js#L575

gabegorelick:
@pokoli Those are the errors from https://github.com/mholt/PapaParse/blob/9155dd5b747cd659a41e0fe760148ee71f89a4e4/papaparse.js#L1156 which is separate from the thrown errors. Future work could make those errors too instead of plain objects.

As far as I can tell the thrown errors are never tested. E.g. there are no instances of `Comment character same as delimiter`, `Input must be a string`, or `Not implemented` that I can find in the tests.

Issue #489, opened 2018-04-13
Skip comments for delimiter detection


candu:
I have CSV files with more than 10 comment lines at the beginning, which trips up `guessDelimiter()` - since that function uses a `Parser` with `preview: 10`, it never sees the data rows.

`guessDelimiter()` should probably respect the `comments:` config option here.

candu:
As a note to anyone else who runs across this, I used `beforeFirstChunk()` to work around this issue:

```
    // rest of config...
    beforeFirstChunk(chunk) {
        let i = 0;
        while (i < chunk.length && chunk[i] === commentPrefix) {
          while (i < chunk.length && !isLineEndingCharacter(chunk[i])) {
            i++;
          }
          while (i < chunk.length && isLineEndingCharacter(chunk[i])) {
            i++;
          }
        }
        return chunk.slice(i);
      }
```

pokoli:
A pull request to skip comments on guest delimiters will be very welcome.

pokoli:
Fixed in #503

PR #503, opened 2018-05-05
When guessing a delimiter, lines with comments are ignored now

efossas:
This PR addresses the issue: "Skip comments for delimiter detection #489".

The guessDelimiter function now takes the comments config and that seems to fix the problem of it failing when there are 10 or more lines of comments at the beginning of CSV.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -898,7 +898,7 @@
 			_delimiterError = false;
 			if (!_config.delimiter)
 			{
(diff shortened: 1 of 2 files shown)


pokoli:
Thanks for PR. 

It will be great if you can use imperative commit message to explain what you are changing. See:

https://chris.beams.io/posts/git-commit/#imperative

This makes history easier to understand. 

Meanwhile I've edited your commit messages so you can see which may be the expected commit message.

Issue #487, opened 2018-04-11
dynamicTyping should use JSON.parse and config should allow custom processing


efossas:
I think dynamicTyping should use JSON.parse to convert data to javascript types. It makes the most sense as its based on a standard (rfc7159) and javascript types are what are usable here anyways.

Since some users might need special conversion (such as converting "True"/"False" to boolean or empty strings to null), the config should have the option to accept a function that applies custom processing to the data.

I have already coded these changes and features (view my fork), as well as made doc changes to the gh-pages branch. Please let me know what you think and if you'll accept a PR with these edits. Thanks.

efossas:
I generated csv strings with 100,000 lines and 8 columns, filled with random strings, booleans, arrays, etc.

These are the results I saw in seconds:

Original PapaParse w/ dynamicTypic on: ~0.70
My updates w/ transform function identical to original dynamicTyping: ~0.75
My updates w/ dynamicTyping using JSON.parse: ~10.6
Replacing JSON.parse w/ json_parse in my updates: ~4.1

The json_parse function can be found at:
https://github.com/douglascrockford/JSON-js/blob/master/json_parse.js

Btw, I noticed a bug in my update where objects are converted to [Object]. This isn't the json parsing functions, it happens somewhere else in PapaParse. I'll look into it later.

At a minimum, I think the ability to add a custom transform function should be added. It does nothing to performance when set to false and gives users more control.

As for using json_parse, what do you think? After some thought, I'm leaning toward removing dynamicTyping all together and just adding examples to the docs on different ways to add dynamicTyping functionality using a custom transform function.

pokoli:
I'm closing this issue as it has been fixed by #505

PR #505, opened 2018-05-05
Add config option that applies custom function to values

efossas:
This PR addresses issue: "dynamicTyping should use JSON.parse and config should allow custom processing #487"

The config can now take the key 'transform' with a function value. That function can apply transformations to the values in a CSV.

This allows the use of the JSON.parse() functionality I discussed without affecting the current dynamicTyping option.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -203,6 +203,8 @@
 		}
 		_config.dynamicTyping = dynamicTyping;
 
+		_config.transform = isFunction(_config.transform) ? _config.transform : false;
+
 		if (_config.worker && Papa.WORKERS_SUPPORTED)
 		{
 			var w = newWorker();
@@ -976,7 +978,7 @@
 			if (needsHeaderRow())
 				fillHeaderFields();
 
-			return applyHeaderAndDynamicTyping();
+			return applyHeaderAndDynamicTypingAndTransformation();
 		}
 
 		function needsHeaderRow()
@@ -1028,9 +1030,21 @@
 			return value;
 		}
 
-		function applyHeaderAndDynamicTyping()
+		function parseTransform(field, value) {
+			if (_config.transform) {
+				if (_config.transform.length === 1) {
+					return _config.transform(value);
+				} else {
+					return _config.transform(field,value);
+				}
+			} else {
+				return value;
+			}
+		}
+
+		function applyHeaderAndDynamicTypingAndTransformation()
 		{
-			if (!_results || (!_config.header && !_config.dynamicTyping))
+			if (!_results || (!_config.header && !_config.dynamicTyping && !_config.transform))
 				return _results;
 
 			for (var i = 0; i < _results.data.length; i++)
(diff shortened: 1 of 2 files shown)


papaparse.js
```diff
@@ -1028,9 +1030,21 @@
 			return value;
 		}
 
-		function applyHeaderAndDynamicTyping()
+		function parseTransform(field, value) {
+			if (_config.transform) {
+				if (_config.transform.length === 1) {
```
pokoli: Why is required to test the number of arguments? I think the function should always accept the same number of arguments

tests/test-cases.js
```diff
@@ -872,6 +872,66 @@ var PARSE_TESTS = [
...
+				return value;
+			}
+		},
+		expected: {
+			data: [[true,false,"T","F"], [true, false, true, false]],
+			errors: []
+		}
+	},
+	{
+		description: "Custom transform function converts leading zero strings to number and empty strings to NULL",
```
pokoli: I do not think this test is required

tests/test-cases.js
```diff
@@ -872,6 +872,66 @@ var PARSE_TESTS = [
...
+				return value;
+			}
+		},
+		expected: {
+			data: [[1,2,3], [null,"null","undefined"]],
+			errors: []
+		}
+	},
+	{
+		description: "Custom transform function uses cache to determine whether to act on column",
```
pokoli: This test is not required also

tests/test-cases.js
```diff
@@ -872,6 +872,66 @@ var PARSE_TESTS = [
 			errors: []
 		}
 	},
+	{
+		description: "Custom transform function converts all true and false strings to booleans regardless of case",
```
pokoli: I will prefer to have a generic test description that talks about that a transform function can be applied and does not talk about the type of transformations applied

tests/test-cases.js
```diff
@@ -872,6 +872,66 @@ var PARSE_TESTS = [
...
+		input: '"true","false","T","F"\r\n"TRUE","FALSE","True","False"',
+		config: {
+			transform: function(value) {
+				if(typeof value === "string" && value.toLowerCase() === "true") return true;
```
pokoli: Can we have a simpler transform function?

papaparse.js
```diff
@@ -1028,9 +1030,17 @@
 			return value;
 		}
 
-		function applyHeaderAndDynamicTyping()
+		function parseTransform(field, value) {
```
pokoli: Do we really need to create such simple function? It's not easier to use it's code directly? I do not see any other place where this code is reused, which makes me think that there is not need of a function.

pokoli:
I just left on minor code issue, once fixed we are ready for merge. 

We host the documentation on the doc folder of the gh-pages branch. Would you mind adding the doc of this new function in another pull request?

academicsystems:
Just updated the docs in #516 PR. Thanks.

pokoli:
Thanks for your effort on this!

PR #506, opened 2018-05-08
Serialize and deserialize dates using ISO format

pushplay:
Dates serialize to ISO format.  ISO date string deserialize to Dates when `dynamicTyping` is true.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -408,7 +408,7 @@
 			if (typeof str === 'undefined' || str === null)
 				return '';
 
-			if (str.constructor === Date)
+			if (str instanceof Date)
 				return JSON.stringify(str).slice(1, 25);
 
 			str = str.toString().replace(quoteCharRegex, _quoteChar + _quoteChar);
--- a/papaparse.js
+++ b/papaparse.js
@@ -408,6 +408,9 @@
 			if (typeof str === 'undefined' || str === null)
 				return '';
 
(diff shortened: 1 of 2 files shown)


papaparse.js
```diff
@@ -408,6 +408,9 @@
 			if (typeof str === 'undefined' || str === null)
 				return '';
 
+			if (str.constructor === Date)
```
pushplay: This comparison is faster than instanceof but won't work for subclasses of Date.  If someone is weird enough to extend Date they can modify toString() to get their desired behavior.

pokoli: I think I will prefer to not overoptimize and use instanceof and make thinks works as expected.

tests/test-cases.js
```diff
@@ -1422,6 +1431,11 @@ var UNPARSE_TESTS = [
 		input: [{"Col1": "a", "Col2": "b", "Col3": "c"}, {"Col1": "d", "Col2": "e", "Col3": "f"}],
 		config: {header: false},
 		expected: 'a,b,c\r\nd,e,f'
+	},
+	{
+		description: "Date handling",
```
pokoli: I think this description can be improved

pokoli:
I've fixed conflics and test description and merged this manually on 757dc379151580de9fd1022254ceadb4b7944d6b

Thanks for your contribution.

Issue #432, opened 2017-11-08
Wrong row reported for `InvalidQuotes`


tarossi:
With version 4.3.6, in some cases, the wrong row is reported. A test case:
```
Papa.parse(
  `a,b,c
   1,2,3,
   4,"5"5",6
   7,8,9`, 
{header:true}).errors
```
Results in
```
[
  {"type":"Quotes","code":"InvalidQuotes","message":"Trailing quote on quoted field is malformed","row":2,"index":16},
  {"type":"FieldMismatch","code":"TooManyFields","message":"Too many fields: expected 3 fields but parsed 4","row":0}
]
```
A field mismatch occurs on row 0 and then one expects the invalid quotes to be reported on row 1 but it's on row 2 instead. This doesn't happen with version 4.3.5 which doesn't report the invalid quotes at all.

Here's the fiddle https://jsfiddle.net/011ev54k/

pokoli:
Thanks for reporting it. 

I think we can add a test case with the corrected behaviour and provide a fix for it.

lunatixcoder:
I think this might be related to 430 which is why its not reporting it as an error any more.  The reason its not an error is that the quotes should now be considered part of the data in the field parsed.  This should be expected behavior.  So I dont believe this is an issue any more.

pokoli:
I can confirm the issue is still present on 4.3.7 (which I just released and deployed on demo site)

PR #509, opened 2018-05-19
Add row counter for chunked parsing for accurate row reporting on errors

mauriciovillalobos:
Add test and long enough csv file to need more than one chunk to process

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -957,7 +957,7 @@
 
 		var self = this;
 		var _stepCounter = 0;	// Number of times step was called (number of rows parsed)
-		var _rowsCounter = 0;	// Total number of rows that have been parsed so far
+		var _rowCounter = 0;	// Number of rows that have been parsed so far
 		var _input;				// The input being parsed
 		var _parser;			// The core parser being used
 		var _paused = false;	// Whether we are paused or not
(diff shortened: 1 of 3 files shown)


tests/node-tests.js
```diff
@@ -59,6 +59,43 @@ describe('PapaParse', function() {
...
+						"message": "Too few fields: expected 3 fields but parsed 2",
+						"row": 498
+					},
+					{
+						"type": "FieldMismatch",
+						"code": "TooFewFields",
+						"message": "Too few fields: expected 3 fields but parsed 2",
+						"row": 998
+					},
+					{
```
pokoli: Is not enought to test with 1000 rows?

papaparse.js
```diff
@@ -957,6 +957,7 @@
 
 		var self = this;
 		var _stepCounter = 0;	// Number of times step was called (number of rows parsed)
+		var _rowsCounter = 0;	// Total number of rows that have been parsed so far
```
pokoli: I think it's better named _rowCounter

pokoli: I think the total word is not requried on the comment, so it will be: _Number of rows that had been parsed so far_

mauriciovillalobos:
Hey,

I updated the PR with your comments

I checked #441 and #432 but this doesn't seem like it's addressing any of those.
The specific issue I was having was on very large csv files, that were being loaded in chunks and the total number of rows parsed was not kept between the chunks.

I think the issue described on #441 has to do with the fact that we're using 0th index to report the row number, and not taking into account the header row if any. This could be fixed by:

- Adding a +1 to offset the 0th index
- Adding another +1 if it has a header row
- Checking if we have skipped empty rows, and take it into account (in the case of empty rows between rows with content)

But this would break current behavior, if anyone was done the offsetting themselves in the code (like I did when working with papaparse just now, I added +2 before reporting the error to my users)

My two cents, this could be fixed but maybe on another PR.

pokoli:
Ok, now I fully understand the picture. I just asked about other issues to make see if they are related but I see is completely unrelated. Thanks for your work on this.

PR #519, opened 2018-06-08
Allow multi-characters delimiter at serialization.

adiroiban:
Scope
=====

This updates the serialization code to allow multi-character delimiters.

I want to present the values in a more human readable format with `, ` as delimiter so that the lines will be `1, 2, 3` instead of `1,2,3`

Changes
========

Update the validation to not enforce single characters, but to still filter against invalid characters.

In this way there is no regression and all previous tests pass.

I added a new test to showcase this functionality

As a drive-by fix, I have updated the phantomjs test runner to propagate the exit code
as it was always exiting with 0 even on errors see https://travis-ci.org/mholt/PapaParse/jobs/389825011#L1418

----------

Once this is accepted the documentation needs to be updated at https://github.com/mholt/PapaParse/blob/gh-pages/docs.html#L256

to inform that it accepts multiple characters

Thanks for the nice library.

I hope you can accept this change. Let me know if you want anything done in another way.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -271,7 +271,7 @@
 		/** whether to write headers */
 		var _writeHeader = true;
 
-		/** delimiting character */
+		/** delimiting character(s) */
 		var _delimiter = ',';
 
 		/** newline character(s) */
--- a/papaparse.js
+++ b/papaparse.js
@@ -326,9 +326,10 @@
 				return;
 
 			if (typeof _config.delimiter === 'string'
-				&& _config.delimiter.length === 1
-				&& Papa.BAD_DELIMITERS.indexOf(_config.delimiter) === -1)
+                && !Papa.BAD_DELIMITERS.filter(function(value) { return _config.delimiter.includes(value); }).length)
 			{
+				// We have a string as config and it does not contains any
+				// of the invalid delimiters.
 				_delimiter = _config.delimiter;
 			}
 
--- a/papaparse.js
+++ b/papaparse.js
@@ -328,8 +328,6 @@
 			if (typeof _config.delimiter === 'string'
                 && !Papa.BAD_DELIMITERS.filter(function(value) { return _config.delimiter.indexOf(value) !== -1; }).length)
 			{
-				// We have a string as config and it does not contains any
-				// of the invalid delimiters.
 				_delimiter = _config.delimiter;
 			}
 
--- a/papaparse.js
+++ b/papaparse.js
@@ -326,7 +326,7 @@
 				return;
 
 			if (typeof _config.delimiter === 'string'
-                && !Papa.BAD_DELIMITERS.filter(function(value) { return _config.delimiter.includes(value); }).length)
+                && !Papa.BAD_DELIMITERS.filter(function(value) { return _config.delimiter.indexOf(value) !== -1; }).length)
 			{
 				// We have a string as config and it does not contains any
 				// of the invalid delimiters.
--- a/tests/test.js
+++ b/tests/test.js
@@ -8,9 +8,8 @@ var server = connect().use(serveStatic(path.join(__dirname, '/..'))).listen(8071
 	if (process.argv.indexOf('--phantomjs') !== -1) {
 		childProcess.spawn('node_modules/.bin/mocha-phantomjs', ['http://localhost:8071/tests/tests.html'], {
 			stdio: 'inherit'
-		}).on('exit', function(code) {
+		}).on('exit', function() {
 			server.close();
-			process.exit(code);  // eslint-disable-line no-process-exit
 		});
 
 	} else {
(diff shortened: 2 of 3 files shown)


adiroiban:
I see that Travis-CI is happy but I now see that phantom fails with
``
TypeError: 'undefined' is not a function (evaluating '_config.delimiter.includes(value)')
```

File: tests/test.js [hunk 1]
```diff
@@ -8,8 +8,9 @@ var server = connect().use(serveStatic(path.join(__dirname, '/..'))).listen(8071
 	if (process.argv.indexOf('--phantomjs') !== -1) {
 		childProcess.spawn('node_modules/.bin/mocha-phantomjs', ['http://localhost:8071/tests/tests.html'], {
 			stdio: 'inherit'
-		}).on('exit', function() {
+		}).on('exit', function(code) {
```
pokoli: Why is this required? It seems not related to the PR

papaparse.js
```diff
@@ -326,9 +326,10 @@
 				return;
 
 			if (typeof _config.delimiter === 'string'
-				&& _config.delimiter.length === 1
-				&& Papa.BAD_DELIMITERS.indexOf(_config.delimiter) === -1)
+                && !Papa.BAD_DELIMITERS.filter(function(value) { return _config.delimiter.indexOf(value) !== -1; }).length)
 			{
+				// We have a string as config and it does not contains any
```
pokoli: Is really needed to add this comment?

tests/test.js
adiroiban: See https://travis-ci.org/mholt/PapaParse/jobs/389825011#L1418

While working on this I commited a bad code and the tests were still green.

So I have updated this to see the tests failing and then fixed the code.

I tried to document it in "Changes"

```
As a drive-by fix, I have updated the phantomjs test runner to propagate the exit code
as it was always exiting with 0 even on errors see https://travis-ci.org/mholt/PapaParse/jobs/389825011#L1418
```

Let me know if the PR description does not make sense and I will try to repharse it.

Thanks!

papaparse.js
adiroiban: nope. is not needed...but that whole logic is pretty long and I wanted to documented the purpose of all those tests

tests/test.js
pokoli: Uops, this sounds as a separate issue which scares me a little bit. 
Could you please include this CHANGE in a separate PR? I will like to merge it first.

papaparse.js
pokoli: For me the code is self explanatory and do not add something relevant. So it can be removed.

tests/test.js
adiroiban: Thanks a lot for the review.

I am not a day to say JS developers so I don't see why this change is so dangerous. 
Why it scares you?

For me, this PR has all the parts for this change and the tests are green.

There is a commit in which we see that the tests are green even with broken code.
There is a separate commit in which the infrastructure code is fixed.
Then we have another commit with bad code which pass on node but fails on phantomjs and for this we can see that Travis-CI is red.
And then a final commit where the phantomjs code is fixed and Travis-CI is fixed.

Putting this into a separate PR would mean that I would have to invent bogus code to have to test break, and then fix the test runner then fix the code... it feels like a lot of extra work.

This is  Boyscout rule https://deviq.com/boy-scout-rule/.

I think that if you put too much extra red tape on code cleanup you might end up with discouraging the cleanup.

I understand the purpose of small PR and a single change in the PR, but I don't think that they should apply to the testing infrastructure code.
I think that you should be allowed to do multiple changes to the infrastrucrure in order to best server the targeted change.

Let me know if you still want this change in a separate and I will do it :)

Thanks again for the review.

pokoli: It scares me that we can find some error that was silently hidden by this code.
Just include the same change in a separate PR, there is no need to add something else. 

I prefer to have a separate PR to keep history cleaner by not include two unrelated changes in the same commit as I normally squash the PR into a single commit before merging.

adiroiban: Done https://github.com/mholt/PapaParse/pull/523

Well, it can happen. There is no test for the test infrastructure so it relies on people writing bad code accidentally (like in my case) or on purpose (if you are religious about tdd )

pokoli:
I think we should mimic the same behaviour when reading a file so a multicharacter file generated with PapaParse can be read by PapaParse. 

Not sure if it's currently supported but if it is we should add a test case for it.

adiroiban:
> Not sure if it's currently supported but if it is we should add a test case for it.

I think is supported. I see https://github.com/mholt/PapaParse/blob/master/papaparse.js#L1348 so I guess that the code takes... or should take into consideration multi-char delimiters.

My use case is handling CSV which is manually created... so I want to be flexible with input parsing/decoding and human readable with unparse/encoding

pokoli:
I've checked and there is a test for it. 

Your pull request is merged, thanks for contributing!

PR #521, opened 2018-06-09
Strip spaces from quoted values.

adiroiban:
Scope
=====

When you have a value which is enclosed in quotes, the spaces outside of the quotes are stripped.

If they are not stripped, papaparse, as of master branch breaks.

But even if it does not break, if you have spaces it return a value like `<space>"B"<space>` (not sure how to add spaces in mardown.
With this, is hard to tell whever that value came from `, "B" ,` or from `," ""B"" ",`

Changes
=======

I have updated an existing test so this introduced a backward incompatibility...but I hope the previous behaviour, even if explicitly documented by a test can be considered a bug.

For the part in which spaces are ignored, I just went with a simple `+1` instead of trying to do anothe search and have a smart cursor positioning

Let me know if you find this useful or you want something changed.
I hope this can be done without adding an extra configuration option.

Thanks!

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -1393,10 +1393,19 @@
 			var nextNewline = input.indexOf(newline, cursor);
 			var quoteCharRegex = new RegExp(escapeChar.replace(/[-[\]/{}()*+?.\\^$|]/g, '\\$&') + quoteChar, 'g');
 			var quoteSearch;
+			var trailingSpacesRegex = new RegExp('^[ ]+(".*)');
+			var trailingSpacesSearch;
 
 			// Parser loop
 			for (;;)
 			{
+				trailingSpacesSearch = input.substring(cursor).match(trailingSpacesRegex);
+				if (trailingSpacesSearch) {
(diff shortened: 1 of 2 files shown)


pokoli:
I'm wondering if we should not add a configuration parameter (disabled by default) to enable this behaviour.

adiroiban:
Thanks for the review.

> I'm wondering if we should not add a configuration parameter (disabled by default) to enable this behaviour.

I am -1 for adding more toggle buttons unless this is really needed.

My main problem with the current (master) behaviour is that you can't  tell whether the value came from ` "B" ` or from `" ""B"" "`

What damaget you think that the behaviour from this PR might do?

pokoli:
This PR changes the current behaviour to adapt the library to a specific use case and that's why I prefer to have a documented option to enable or disable it. 

This allows any user of the library to decide which behaviour it prefers which for me is one of the main objectives of a generic library for parsing CSV files. 

What's wrong with the configuration option?

pokoli:
I'm closing this as this is now supported by the transform headers functionn.

PR #534, opened 2018-07-06
Fixed #532 - Implemented onlyQuoteStrings

### LINKED ISSUE #532: Quotedifferent data types
I'm parsing the following file (quoted strings, non-quotes numbers):

```
name,age,city
"Chirstian",30,"Hamburg"
```

The option `dynamicTyping` detects name and city as strings and the age as number. If I `unparse` my data, I have two options.

##### Option 1) set `quoteChar: '"'`, which results in:

```
name,age,city
"Chirstian","30","Hamburg"
```

##### Option 1) Skip the option `quoteChar, which results in:

```
name,age,city
Chirstian,30,Hamburg
```

I recommend an option `onlyQuoteStrings`, which applies the `quoteChar: '"'` only to strings, but not to numbers, similar to libre office save as CSV dialogue (option `quote all text cells`):

bildschirmfoto vom 2018-07-05 23 59 54

boldt:


papaparse.js
```diff
@@ -412,9 +420,15 @@
 			if (str.constructor === Date)
 				return JSON.stringify(str).slice(1, 25);
 
+			// We must check, whether the data type is string already here,
+			// since all data types are converted into strings in the next line
+			var onlyQuoteStrings = _onlyQuoteStrings && typeof str === 'string';
```
pokoli: I think we should rename str into value here to make code clearer.

boldt:
* I renamed `str` into `value`
* I added a documentation for the new property
* I also formatted the unparse config as an unordered list
    * Maybe it maes sense, to format the `unparse` options [the same stle as the parse options}(https://www.papaparse.com/docs#config)

pokoli:
After reading the docs I'm wondering if we should not allow three options on the quotes parameters:
- false
- true
- 'only-strings'

With a second though I'm wondering if we should allow to pass a function to the quotes argument which will accept the value and the col and return true if the field needs quoting. This will prevent adding more settings like: 'only-numbers' and give the full power to the end user. For the string use case it will be:
```
config: {
'quotes': function(value, col) { return typeof value == 'string' },
}
```
What do you think?

boldt:
I think, adding a function is a quite nice idea! This makes it much more flexible. 

I would like to keep the true/false to be backward compatible, thus upgrading to a newer version does not break existing applications (e.g., true/false will be internally implemented as a function as well), thus:

* `false` - do not add quotes (default)

```
'quotes': function(value, col) { return false; }
```

* `true` - add quotes

```
'quotes': function(value, col) { return false; }
```

* a function is provided - use the function (like your example)

pokoli:
Your proposal makes sense for me. Please go for the implementation!

Sector7B:
> @boldt Are you still planning on implementing the function option? I can help out if you want.

Much needed feature for our team if you're looking for any motivation.  ;-)

PR #449, opened 2018-01-27
Fixing regex from pokoli, updating test

karllhughes:
Addresses #252 by fixing some little issues with @pokoli's PR: https://github.com/mholt/PapaParse/compare/master...newline-problem

I simply edited his fix, so if appropriate, I can make the PR on his branch instead? I just wasn't sure the best flow for this kind of thing. Thanks!

PR #542, opened 2018-07-24
Fix linebreak problem in headers

### LINKED ISSUE #252: Error parsing file when CR appears in header in quotes
Hi Matt, thanks for the awesome library!

I have encountered a potential issue regarding newline characters, and it would be great to hear your thoughts on it. 

Some files I'm using rather unusually have CR characters in the headers, although they appear in quotes. The end of the line is a normal LF character. I think when PapaParse tries to automatically detect the newline characters, it will use CR which leads to parsing errors (MissingQuotes). Obviously I could tell PapaParse the newline character, but ideally I'd like to continue using the automatic detection because I don't always know what it's going to be.

Would it be possible to ignore newline characters which are in quotes when trying to guess the CSV end of line character?

Here is an example file which exhibits the problem: https://www.dropbox.com/s/d6qkvywinjekulh/header-newline.csv?dl=0

Thanks!

### LINKED ISSUE #457: Unable to parse files with \n and \r\n
I am having a little issue with PapaParse when parsing a file that contains newlines in fields. In particular, the file has `\r\n` as line separator, but inside fields there are `\n` as new lines.

When I load such files, PapaParse recognizes `\n` as newline delimiter instead of `\r\n`, so the entire file is read in the first row.

An example file follows, base64-encoded (there is BOM at the beginning).

```
77u/Zmlyc3Qtcm93LCJMaW5lICIid2l0aCBxdW90ZWQiIiB0ZXh0CgpGb28gYmFyIGJheiINCnNl
Y29uZC1yb3csInNvbWUgb3RoZXIgIiJxdW90ZWQiIiB0ZXh0IAoKTGluZSBhZnRlciBlbXB0eSIN
Cmxhc3QtZW1wdHktbGluZSw=
```

In particular, when I try parsing the file I get

```
// ff is the file string above
Papa.parse(atob(ff)).data.length // 1
JSON.stringify(Papa.parse(atob(ff)))
"{\"data\":[[\"ï»¿first-row\",\"Line \\\"\\\"with quoted\\\"\\\" text\\n\\nFoo bar baz\\\"\\r\\nsecond-row,\\\"some other \\\"\\\"quoted\\\"\\\" text \\n\\nLine after empty\\\"\\r\\nlast-empty-line,\"]],\"errors\":[{\"type\":\"Quotes\",\"code\":\"MissingQuotes\",\"message\":\"Quoted field unterminated\",\"row\":0,\"index\":14}],\"meta\":{\"delimiter\":\",\",\"linebreak\":\"\\n\",\"aborted\":false,\"truncated\":false,\"cursor\":131}}"
```

I don't know what delimiters are used inside fields and between records, so I am unable to specify the `newline` option when parsing. As a reference, LibreOffice is parsing the above file correctly.

Am I missing some option or is this a bug?
Thanks!

### LINKED ISSUE #535: InvalidQuotes error only when parsing file directly
Please see the attached example CSV file. Using this local file on https://www.papaparse.com/demo produces a "InvalidQuotes" error. However, if I open the file in a text editor and simply copy/paste the same exact data directly to the demo it works without error.

[upm.zip](https://github.com/mholt/PapaParse/files/2182902/upm.zip)

File data

```
Google,googleuser,0UI9eH4$,google.com,"This is a note

Line1
line2"
Yahoo,,k7F3EKZ>,,
GitHub,githubuser,~UAWk2DO,http://github.com,
```

MonkeyDZeke:
Supersedes https://github.com/mholt/PapaParse/pull/449 with improved regex and added tests.
Resolves #252 
Resolves #457
Resolves #535 

Test-driven correction for bugs regarding linebreaks in quoted header fields that are not the same time as the linebreaks between rows. I was able to mostly fix it using an improved regex (using `[^]*?` instead of `.*?` for replacing across newlines).

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -1257,7 +1257,7 @@
 		{
 			input = input.substr(0, 1024 * 1024);	// max length 1 MB
 			// Replace all the text inside quotes
-			var re = new RegExp(quoteChar + '([^]*?)' + quoteChar, 'gm');
+			var re = new RegExp(escapeRegExp(quoteChar) + '([^]*?)' + escapeRegExp(quoteChar), 'gm');
 			input = input.replace(re, '');
 
 			var r = input.split('\r');
@@ -1290,9 +1290,10 @@
 		}
 	}
 
-
-
-
+	/** https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions */
+	function escapeRegExp(string) {
+		return string.replace(/[.*+?^${}()|[\]\\]/g, '\\$&'); // $& means the whole matched string
+	}
 
 	/** The core parser implements speedy and correct CSV parsing */
 	function Parser(config)
--- a/papaparse.js
+++ b/papaparse.js
@@ -1008,7 +1008,7 @@
 		 */
 		this.parse = function(input, baseIndex, ignoreLastRow)
 		{
-            var quoteChar = _config.quoteChar || '"';
+			var quoteChar = _config.quoteChar || '"';
 			if (!_config.newline)
 				_config.newline = guessLineEndings(input, quoteChar);
 
@@ -1258,7 +1258,7 @@
 			input = input.substr(0, 1024 * 1024);	// max length 1 MB
 			// Replace all the text inside quotes
 			var re = new RegExp(quoteChar + '(.*?)' + quoteChar, 'gm');
-			input = input.replace(re, '')
+			input = input.replace(re, '');
 
 			var r = input.split('\r');
 
--- a/papaparse.js
+++ b/papaparse.js
@@ -1008,8 +1008,9 @@
 		 */
 		this.parse = function(input, baseIndex, ignoreLastRow)
 		{
+            var quoteChar = _config.quoteChar || '"';
 			if (!_config.newline)
-				_config.newline = guessLineEndings(input);
+				_config.newline = guessLineEndings(input, quoteChar);
 
 			_delimiterError = false;
 			if (!_config.delimiter)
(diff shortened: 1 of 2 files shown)


MonkeyDZeke:
I based this off of this PR https://github.com/mholt/PapaParse/pull/449 and tried to update it to the current state of the repo. Might have messed things up, though. Any suggestions on resolving the conflicts?

papaparse.js
```diff
@@ -1252,9 +1253,12 @@
 			};
 		}
 
-		function guessLineEndings(input)
+		function guessLineEndings(input, quoteChar)
 		{
 			input = input.substr(0, 1024 * 1024);	// max length 1 MB
+			// Replace all the text inside quotes
+			var re = new RegExp(quoteChar + '([^]*?)' + quoteChar, 'gm');
```
dboskovic: The `quoteChar` here should be run through a regex escape. It could cause major issues if the user provided a character like `.` as the quote char.

This function from the mozilla docs should be good if there isn't already one available in this repo. (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions)
```js
function escapeRegExp(string) {
  return string.replace(/[.*+?^${}()|[\]\\]/g, '\\$&'); // $& means the whole matched string
}
```

tests/test-cases.js
```diff
@@ -1208,13 +1208,131 @@ var PARSE_TESTS = [
 			data: [{'a': 'c', 'b': 'd'}],
 			errors: []
 		}
+	},
```
dboskovic: Probably a test here to verify that using a regexy quote character doesn't break things.

pokoli: I agree that the test will be a good aditions

MonkeyDZeke:
I used the `.` character for the test since I figured that would break things the worst and would demonstrate the problem and the fix best.

pokoli:
Thanks for fixing this long standing issue!

MonkeyDZeke:
My pleasure! It's a fantastic project, and I'm proud to contribute to it. :slightly_smiling_face:

PR #540, opened 2018-07-20
Add config option to convert headers to lowercase

ntelkedzhiev:
Allows you to convert headers to lowercase which is particularly useful when parsing CSV files prone to human errors.

PR #543, opened 2018-07-25
Add option to strip headers out of white spaces

ntelkedzhiev:
Remove all whitespaces in headers.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -1113,8 +1113,8 @@
 						header = header.trim();
 					}
 
-					if (_config.stripHeaders) {
-						header = header.replace(/\s/g,'');
+					if (_config.headerTransform) {
+						header = _config.headerTransform(header);
 					}
 
 					_fields.push(header);
--- a/papaparse.js
+++ b/papaparse.js
@@ -1118,7 +1118,7 @@
 							header = _config.headerTransform(header);
 						} else {
 							throw 'headerTransform must be a function';
-						}	
+						}
 					}
 
 					_fields.push(header);
--- a/papaparse.js
+++ b/papaparse.js
@@ -1113,8 +1113,12 @@
 						header = header.trim();
 					}
 
-					if (typeof _config.headerTransform === 'function') {
-						header = _config.headerTransform(header);
+					if (_config.headerTransform) {
+						if (typeof _config.headerTransform === 'function') {
+							header = _config.headerTransform(header);
+						} else {
+							throw 'headerTransform must be a function';
+						}	
 					}
 
 					_fields.push(header);
--- a/papaparse.js
+++ b/papaparse.js
@@ -1113,6 +1113,10 @@
 						header = header.trim();
 					}
 
+					if (_config.stripHeaders) {
+						header = header.replace(/\s/g,'');
+					}
+
 					_fields.push(header);
 				}
 			_results.data.splice(0, 1);
--- a/papaparse.js
+++ b/papaparse.js
@@ -1113,7 +1113,7 @@
 						header = header.trim();
 					}
 
-					if (_config.headerTransform) {
+					if (typeof _config.headerTransform === 'function') {
 						header = _config.headerTransform(header);
 					}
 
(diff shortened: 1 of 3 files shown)


tests/test-cases.js
```diff
@@ -733,6 +733,15 @@ var PARSE_TESTS = [
 			errors: []
 		}
 	},
+	{
+		description: "Header rows are striped out of white spaces when stripHeaders is set ",
+		input: '  A , B  ,  C  \r\na,b,c',
```
dboskovic: This tests the same thing as `trimHeaders`. You should write a test that actually demonstrates the unique function of this which is to remove all white space from a header entirely. Is your intention to convert a header like `full name` to `fullname`?

dboskovic:
Overall this seems like a really opinionated feature to add to Papaparse. `trimHeaders` takes care of the left and right white space issue which is an obvious data integrity issue. This transforms the actual header values in a pretty subjective way.

I think maybe a better way to build this particular feature is to pass a callback function that can be run on header values. Something like:

```js
{
  ...
  headerTransformer: function(header) {
     return _.dasherize(header)
  }
}
```

ntelkedzhiev:
Thanks for the great suggestion! If a header transformer function is added, do you think that removing`trimHeaders` will be a good idea?

dboskovic:
I think that `trimHeaders` makes sense as a core feature because its purpose is to return the _intended_ data.

In contrast something like removing whitespace in the middle of a text block is more about transforming data into something entirely different than the original data so it makes sense as a more abstract function that can allow for different opinions. (ie. I would usually dasherize or camelCase my headers rather than just strip whitespace.)

ntelkedzhiev:
Okay, makes sense. Thanks for the input. 

Added a headerTransform function and closing my other PR - #540.

papaparse.js
```diff
@@ -1113,8 +1113,8 @@
 						header = header.trim();
 					}
 
-					if (_config.stripHeaders) {
-						header = header.replace(/\s/g,'');
+					if (_config.headerTransform) {
```
dboskovic: Probably make this check for callability rather than truthiness? `typeof _config.headerTransform === 'function'`

dboskovic:
Probably update the docs and add the new config option here:
https://github.com/mholt/PapaParse/blob/master/docs/docs.html#L358

and here
https://github.com/mholt/PapaParse/blob/master/docs/docs.html#L545

papaparse.js
```diff
@@ -1113,6 +1113,10 @@
 						header = header.trim();
 					}
 
+					if (typeof _config.headerTransform === 'function') {
```
pokoli: HeaderTransform should always be a function so instead of silently ignoring the arugment if it is not I prefer to make PapaParse throw and error when it is not a function.

pokoli:
Travis is failing, could you please fix it before merging?

pokoli:
I think it's better to remove the trimHeaders option and in favour of the transformHeaders by passing the trim function. 

What you you think?

dboskovic:
@pokoli I would agree if this was the first implementation but I don’t think it’s necessary to make this a v5 feature by adding a breaking change.

pokoli:
I prefer to have only a single way of achiving something so for me it's worth to have a breaking change. 

BTW, we already have two other PR with any API change, so we can release 4.6 with the compatible features, and then just include the breaking on master. 

If you are worried about supporting bugs on 4.6 release, we can create a branch to include some potential bug fixes for a while in the last stable version. 

Maybe we need a Relase plan?

ivosabev:
Is this going to be merged soon?

pokoli:
I will prefer to make the API change instead of merging it as is

ivosabev:
If you outline your idea we can rework the pr

pokoli:
@ivosabev see https://github.com/mholt/PapaParse/pull/543#issuecomment-409240035

pokoli:
I'm closing this as #589 has been merged to master (which is 5.0 now and is unreleased). 

I also created the 4.x branch if you want to include this feature on the released version (without the API change) feel free to create a PR which targets to the 4.x branch

PR #544, opened 2018-07-26
New feature: Option to skip lines with no content

MonkeyDZeke:
Added a boolean config value `skipNoContentLines` which skips lines that only contain delimiters, quotes, and whitespace, instead of just lines that evaluate to an empty string. Works if set alongside `skipEmptyLines` or on its own. Passing tests are included.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -1015,7 +1015,7 @@
 			_delimiterError = false;
 			if (!_config.delimiter)
 			{
-				var delimGuess = guessDelimiter(input, _config.newline, (_config.skipEmptyLines || _config.skipNoContentLines), _config.comments);
+				var delimGuess = guessDelimiter(input, _config.newline, _config.skipEmptyLines, _config.comments);
 				if (delimGuess.successful)
 					_config.delimiter = delimGuess.bestDelimiter;
 				else
@@ -1077,7 +1077,7 @@
 
 		function testEmptyLine(s, d) {
 			var q = _config.quoteChar || '"';
-			var r = _config.skipNoContentLines ? new RegExp('^[' + escapeRegExp(d) + escapeRegExp(q) + '\\s]*$') : new RegExp('^$');
+			var r = _config.skipEmptyLines === 'strict' ? new RegExp('^[' + escapeRegExp(d) + escapeRegExp(q) + '\\s]*$') : new RegExp('^$');
 			for (var i = 0; i < s.length; i++)
 				if (r.test(s[i]))
 					return true;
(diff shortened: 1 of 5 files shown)


papaparse.js
```diff
@@ -1076,7 +1076,9 @@
 
 		function testEmptyLine(s, d) {
 			var q = _config.quoteChar || '"';
-			var r = _config.skipNoContentLines ? new RegExp('^[' + d + q + '\\s]*$') : new RegExp('^$');
+			var r = _config.skipNoContentLines ? new RegExp('^[' + escapeRegExp(d) + escapeRegExp(q) + '\\s]*$') : new RegExp('^$');
+			// var q = _config.quoteChar || '"';
```
pokoli: Please do not include commented lines

MonkeyDZeke:
That's also an option. Whatever is most maintainable and user-friendly. We could use 'greedy', 'strict', 'no-content', or a few different options. Whatever we do, the documentation needs to be super clear and intuitive.

MonkeyDZeke:
Having the mixed string and boolean input makes the demo for it a bit awkward so I removed that for now until we decide if we want to demo this particular feature and how it would be best done.

papaparse.js
```diff
@@ -1086,7 +1095,7 @@
 			if (_config.skipEmptyLines)
 			{
 				for (var i = 0; i < _results.data.length; i++)
-					if (_results.data[i].length === 1 && _results.data[i][0] === '')
+					if (testEmptyLine(_results.data[i], _config.delimiter))
```
dboskovic: So, I feel like, if we do it this way, since it's post parsing you only need to test that all values are empty in the data array. If you use the regex, then it should be on the raw content if it's available.

MonkeyDZeke: So do we want to let through values that have whitespace, delimiters that were escaped in a quoted field, escaped quotes, and that kind of flotsam? So like if we do something (in ES5 of course) like this:

```
if (_config.skipEmptyLines) {
  _results.data = _results.data.filter(row => row.some(cell => cell !== ''))
}
```

It would behave very differently from what the current feature does.

dboskovic: Interesting, we should stripe whitespace in this test, but I think escaped values should not be part of this. `row.join('').trim() === ''` might be the thing to test in greedy mode.

pokoli:
For the demo I think we can preserve the current boolean flag and set keep the previous behaviour.

papaparse.js
```diff
@@ -1290,8 +1295,8 @@
 		}
 	}
 
-	/** https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions */
-	function escapeRegExp(string) {
+	function escapeRegExp(string)
```
pokoli: Why changing this lines?

MonkeyDZeke: Consistency with the coding style of the rest of the functions.

dboskovic: I think the comment referencing the source material makes sense to keep.

PR #546, opened 2018-07-31
XMLHttpRequest compatibility for Node

jlitowitz:
Node doesn't natively support XMLHttpRequest.

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,5 +53,8 @@
 		"test-phantomjs": "node tests/test.js --phantomjs",
 		"test-node": "mocha tests/node-tests.js tests/test-cases.js",
 		"test": "npm run lint && npm run test-node && npm run test-phantomjs"
+	},
+	"dependencies": {
+	  "xmlhttprequest": "^1.8.0"
 	}
 }
--- a/papaparse.js
+++ b/papaparse.js
@@ -4,6 +4,7 @@
 	https://github.com/mholt/PapaParse
 	License: MIT
 */
+var XMLHttpRequest = XMLHttpRequest || require('xmlhttprequest').XMLHttpRequest;
 (function(root, factory)
 {
 	/* globals define */


dboskovic:
I'm not sure how I feel about adding a shim dependency for `XMLHttpRequest` in NodeJS when the `http` and `https` libraries are available. On the other hand the refactor to support `http` requests is significant. 🤔

pokoli:
I also prefer to not add more dependencies on the project. 

Is there any way to get the contents of a http requests as Node Stream? As we support Node Streams, this will work without any extra effort.

jlitowitz:
Well, there is this, which seems to be towards the direction @pokoli is thinking:

[https://nodejs.org/api/http.html#http_agent_createconnection_options_callback](https://nodejs.org/api/http.html#http_agent_createconnection_options_callback
)
> callback <Function> Callback function that receives the created socket

I don't have any experience in using that though.  For my purposes, the one-line commit was good enough; thought I'd share it around.

jlitowitz:
Well, there is this, which I think is towards what @pokoli is thinking:

[https://nodejs.org/api/http.html#http_agent_createconnection_options_callback](https://nodejs.org/api/http.html#http_agent_createconnection_options_callback)

> Produces a socket/stream to be used for HTTP requests.

I don't have any experience using it this though.

dboskovic:
@pokoli do you think we should target parity between NodeJS and browser or start feature-flagging things for Node vs. Browser?

pokoli:
@dboskovic we should diferntiate as thinks do not work the same way on Node that on Browser. 
For me as far as there is a solution that works without any change on the library we should encourage it's usage. 

So for me we should reject this pull request which add s a dependencies that can be achieved by using node streams. 

On the other hand I will be very happy to have a section explaining how to achieve browser functionalities on Node

dboskovic:
While I agree that there's environment difference, I think we advertise NodeJS compatibility and I think it's not a huge leap to add API parity on NodeJS (especially in this case). I think we should also Node ENV unit tests to our CI to represent that.

@jlitowitz I've added an issue for this and referenced this PR. You should be able to follow along when it's available!

PR #554, opened 2018-08-14
Use 'greedy' skipEmptyLines in unparse

### LINKED ISSUE #553: Either skipEmptyLines is broken--or is not used for unparse()
**Expected**
Calling unparse with skipEmptyLines: 'greedy' results in no basically empty rows.

**Actual**
Calling unparse with skipEmptyLines: 'greedy' (or true) will still have rows with no data (except commas). 

**Sample Code**
```
let csv = Papa.unparse(
      {
        fields: ['a', 'b'],
        data: [[null, ' '], [], ['1', '2']]
      },
      {
        skipEmptyLines: 'greedy',
        worker: true
      }
 );
console.log(csv);
```

**Discussion**
Am I misunderstanding something or not setting the config correctly? Is `skipEmptyLines` only for parse, perhaps? In the mean time, I suppose I can do some edge checking so I don't pass "empty" rows to unparse.... :(

Keep up the good work! I love the project. Glad I found it. Fast, clean, modern, reliable.

MonkeyDZeke:
Added support for skipEmptyLines in unparsing using both `true` and `'greedy'` values.

Resolves #553

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -293,9 +293,9 @@
 		if (_input instanceof Array)
 		{
 			if (!_input.length || _input[0] instanceof Array)
-				return serialize(null, _input, _skipEmptyLines === 'greedy');
+				return serialize(null, _input, _skipEmptyLines);
 			else if (typeof _input[0] === 'object')
-				return serialize(objectKeys(_input[0]), _input, _skipEmptyLines === 'greedy');
+				return serialize(objectKeys(_input[0]), _input, _skipEmptyLines);
 		}
 		else if (typeof _input === 'object')
 		{
@@ -316,7 +316,7 @@
 					_input.data = [_input.data];	// handles input like [1,2,3] or ['asdf']
 			}
 
-			return serialize(_input.fields || [], _input.data || [], _skipEmptyLines === 'greedy');
+			return serialize(_input.fields || [], _input.data || [], _skipEmptyLines);
 		}
 
 		// Default (any valid paths should return before this)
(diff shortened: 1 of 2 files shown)


papaparse.js
```diff
@@ -335,6 +338,10 @@
 				|| _config.quotes instanceof Array)
 				_quotes = _config.quotes;
 
+			if (typeof _config.skipEmptyLines === 'boolean'
```
pokoli: Why testing the typeof skipemptyLines? Is his really required?

MonkeyDZeke:
Setting a default value for it is required, or else testing it will throw
an error if it's not set. I followed the coding pattern for that already
established, which includes checking the data type when overriding the
defaults with the config.

pokoli:
@MonkeyDZeke that makes sense. Thanks for the explanation!

MonkeyDZeke:
You're welcome! :)

tests/test-cases.js
```diff
@@ -1636,6 +1636,24 @@ var UNPARSE_TESTS = [
...
+		expected: '," "\r\n\r\n1,2'
+	},
+	{
+		description: "Returns without empty rows when skipEmptyLines is true",
+		input: [[null, ' '], [], ['1', '2']],
+		config: {skipEmptyLines: true},
+		expected: '," "\r\n1,2'
+	},
+	{
+		description: "Returns without rows with no content when skipEmptyLines is 'greedy'",
```
bean5: :+1:

PR #555, opened 2018-08-15
Added additional config option, additionalDelimitersToGuess

ghost:
This seemed easy enough to add in and non-intrusive to anything else, so I figured I would just create a pull request instead of opening up a feature request.

This adds an additional config option called `additionalDelimitersToGuess` where a user can pass an array of additional delimiters for PapaParse to try and guess.

For example:

```
var results = Papa.parse('"A"~"B"~"C"', {additionalDelimitersToGuess: ['~']);
results.meta.delimiter // "~"
```

I have included a couple additional tests. Please let me know if you would like more.

### DIFF
--- a/docs/docs.html
+++ b/docs/docs.html
@@ -356,7 +356,7 @@ var csv = Papa.unparse({
 	beforeFirstChunk: undefined,
 	withCredentials: undefined,
 	transform: undefined,
-	delimitersToGuess: [',', '\t', '|', ';', String.fromCharCode(30), String.fromCharCode(31)]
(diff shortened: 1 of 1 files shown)


pokoli:
Thanks for your pull request.

I think we can name it _delimitersToGuess_ and fall back to the default ones if not defined. 

Could you please document the usage of this new config parameter in the docs? They are on the docs folder of this same repository. 

Please do not modify the minified file, this is part of the release process

astupidnerd:
@pokoli, sounds good, I can do that.

astupidnerd:
@pokoli I've updated the pull request with those changes. Let me know if there's anything else I can do.

nosovk:
@pokoli could you please merge it?

Issue #558, opened 2018-08-20
Test case is failing (for skipEmptyLines when header is provided)


bean5:
See also: 
* https://github.com/mholt/PapaParse/issues/553#issuecomment-413287866
* https://github.com/mholt/PapaParse/pull/554

PR #569, opened 2018-09-21
Fixes unparsing using 'skipEmptyLines' = 'greedy' when using headers

MonkeyDZeke:
Resolves https://github.com/mholt/PapaParse/issues/558

See also:

* [#553 (comment)](https://github.com/mholt/PapaParse/issues/553#issuecomment-413287866)
* #554

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -398,7 +398,9 @@
 				var emptyLine = false;
 				var nullLine = hasHeader ? Object.keys(data[row]).length === 0 : data[row].length === 0;
 				if (skipEmptyLines && !hasHeader)
+				{
 					emptyLine = skipEmptyLines === 'greedy' ? data[row].join('').trim() === '' : data[row].length === 1 && data[row][0].length === 0;
+				}
 				if (skipEmptyLines === 'greedy' && hasHeader) {
 					var line = [];
 					for (var c = 0; c < maxCol; c++) {
@@ -416,8 +418,10 @@
 						var colIdx = hasHeader && dataKeyedByField ? fields[col] : col;
 						csv += safe(data[row][colIdx], col);
 					}
-					if (row < data.length - 1 && (!skipEmptyLines || maxCol > 0) && (!skipEmptyLines || !nullLine))
+					if (row < data.length - 1 && (!skipEmptyLines || (maxCol > 0 && !nullLine)))
+					{
 						csv += _newline;
+					}
 				}
 			}
 			return csv;
--- a/papaparse.js
+++ b/papaparse.js
@@ -394,18 +394,29 @@
 			for (var row = 0; row < data.length; row++)
 			{
 				var maxCol = hasHeader ? fields.length : data[row].length;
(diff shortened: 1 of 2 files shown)


papaparse.js
```diff
@@ -394,18 +394,29 @@
...
 					for (var col = 0; col < maxCol; col++)
 					{
-						if (col > 0)
+						if (col > 0 && !nullLine)
 							csv += _delimiter;
 						var colIdx = hasHeader && dataKeyedByField ? fields[col] : col;
 						csv += safe(data[row][colIdx], col);
 					}
-					if (row < data.length - 1 && (!skipEmptyLines || maxCol > 0))
+					if (row < data.length - 1 && (!skipEmptyLines || maxCol > 0) && (!skipEmptyLines || !nullLine))
```
dboskovic: You can rollup that last conditional group into the previous one:

```(!skipEmptyLines || maxCol > 0) && (!skipEmptyLines || !nullLine)```
`===`
```(!skipEmptyLines || maxCol > 0 || !nullLine)```

File: papaparse.js [hunk 2]
```diff
@@ -394,18 +394,29 @@
 			for (var row = 0; row < data.length; row++)
 			{
 				var maxCol = hasHeader ? fields.length : data[row].length;
-				var r = hasHeader ? fields : data[row];
 
-				if (skipEmptyLines !== 'greedy' || r.join('').trim() !== '')
+				var emptyLine = false;
+				var nullLine = hasHeader ? Object.keys(data[row]).length === 0 : data[row].length === 0;
+				if (skipEmptyLines && !hasHeader)
+					emptyLine = skipEmptyLines === 'greedy' ? data[row].join('').trim() === '' : data[row].length === 1 && data[row][0].length === 0;
```
dboskovic: I know this library has a lot of un-braced conditionals. But I'd love to avoid that going forwards. Can you make sure all condition blocks in new code have `{}`?

MonkeyDZeke: Gotcha.

papaparse.js
MonkeyDZeke: Not quite equivalent. I wrote them rolled together like that originally and the tests fail (specifically when `skipEmptyLines` is `true` when unparsing with headers). We have two options as far as I can tell:

`(!skipEmptyLines || maxCol > 0) && (!skipEmptyLines || !nullLine)`

or

`(!skipEmptyLines || (maxCol > 0 && !nullLine))`

 If `skipEmptyLines === true` and `maxCol > 0 === true` then the newline will be added even if `nullLine === true` which is bad. We only want to check `maxCol > 0` and `!nullLine` if `skipEmptyLines === true` but both of them must be true in that case.

I thought the former was more readable, but we can do the latter if you prefer it.

papaparse.js
pokoli: If you want to ensure this behaviour maybe we can include a rule in .eslintrc.js and fix all current dead code ;)

bean5:
I just got around to testing this. It works as expected.

:1st_place_medal:

Issue #235, opened 2015-07-20
Papa.unparse() failing to detect Array of objects


bourquep:
I'm using `BabyParse.unparse()` to convert an array of objects to CSV. My code runs server-side in a [Parse cloud function](https://parse.com/docs/js/guide#cloud-code-cloud-functions).

My code is along those lines:

``` javascript
query.find().then(function(results) {
    var data = [];
    results.forEach(function(r) {
        data.push({
            col1: r.get('field1'),
            col2: r.get('field2'),
        });
    });

    var csv = babyparse.unparse(data);
    // csv is empty!
});
```

I've traced the source of the problem to this part of `BabyParse.JsonToCsv()`:

``` javascript
if (_input instanceof Array) // THIS RETURNS FALSE
{
    if (!_input.length || _input[0] instanceof Array)
        return serialize(null, _input);
    else if (typeof _input[0] === 'object')
        return serialize(objectKeys(_input[0]), _input);
}
else if (typeof _input === 'object')
{
        // It thinks my source array is an object!
}
```

I am no Javascript expert, so I can't comment on the nuances of `typeof` vs `instanceof` vs other methods of checking a type, but [this SO answer](http://stackoverflow.com/a/4775737/114623) seems to say that the preferred way to check if a variable is an array is:

``` javascript
Object.prototype.toString.call( someVar ) === '[object Array]'
```

I patched my copy of `babyparse.js` to use this check instead of `instanceof Array`, and sure enough it works and I get the csv I want.

Anyone who understands these JS intricacies better than I do care to comment?

mholt:
Ugh, that's weird. That change feels so hacky... but I guess it is JavaScript. Will leave this open for a bit to see if anyone else has any thoughts before we go forward.

bluej100:
If we'd rather require IE9 than include the prototype.toString, we could use [Array.isArray](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray).

This comment from that SO question is probably roughly why it's failing: "v instanceof Array will return false if v was created in another frame (v is instance of thatFrame.contentWindow.Array class)."

mholt:
Interesting. After looking into it some more, it seems that toString may be the best way to go about it and most standard among all the engines. I'm fine with that.

bean5:
@bourquep could you open a PR with the change? Are there other places in the code that could adopt the accepted SO advice?

bourquep:
[...]

PR #570, opened 2018-09-21
Object.prototype.toString instead of instanceOf to determine arrayness

MonkeyDZeke:
Resolves https://github.com/mholt/PapaParse/issues/235

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -41,6 +41,14 @@
 		return {};
 	})();
 
+	// Polyfills
+	// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray#Polyfill
+	if (!Array.isArray)
+	{
+		Array.isArray = function(arg) {
+			return Object.prototype.toString.call(arg) === '[object Array]';
+		};
+	}
 
 	var IS_WORKER = !global.document && !!global.postMessage,
 		IS_PAPA_WORKER = IS_WORKER && /(\?|&)papaworker(=|&|$)/.test(global.location.search),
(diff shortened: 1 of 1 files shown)


MonkeyDZeke:
I'm puzzled by that as well. I had thought it might be something to do with my local build since that error shows up on master as well for me. Evidently not. :thinking:

pokoli:
It seems that something changed and our test suite is broken now (all PR are failing).

I think we should fix our test suite before merging any new code.

it seems a problem with the chai.js version [which is loaded from their website ](https://github.com/mholt/PapaParse/blob/master/tests/tests.html#L9). Probably we should include or own version of chai.js with a pinned dependency

pokoli:
Regarding the PR could you please add a test suite that test the isArray function against a basic case and some failing case?

MonkeyDZeke:
I could I suppose, but it's literally a direct copy of the [official polyfill](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray#Polyfill) for `Array.isArray`. And it's hard to test the original failing case because it only fouls up when the array is set in a different frame from where it's tested (from what I understand).

pokoli:
@MonkeyDZeke Then I think we should use the official polyfill when available, and define all needed polyfils on the top of the papaparse.js file.

MonkeyDZeke:
@pokoli Like so?

papaparse.js
```diff
@@ -41,6 +41,14 @@
 		return {};
 	})();
 
+	// Polyfills
```
pokoli: I think we should also define in the parent scope so

MonkeyDZeke:
@pokoli Is that what you were meaning?

papaparse.js
```diff
@@ -1,9 +1,19 @@
 /*@license
-	Papa Parse
-	v4.6.0
-	https://github.com/mholt/PapaParse
-	License: MIT
+Papa Parse
+v4.6.0
+https://github.com/mholt/PapaParse
```
pokoli: why changing license?

pokoli:
@MonkeyDZeke yes, that's what I was looking for

MonkeyDZeke:
Cool, should be good then, once we figure out what's breaking the test suite.

pokoli:
Could you rebase with latest mater branch so we could ensure the test is passing and then merge it? 

Thanks

MonkeyDZeke:
Rebased and ready for merging. :smile:

PR #575, opened 2018-10-01
Bumping font size and darkening color on h5, footer h4

oleersoy:
Bumping font size and darkening color on h5, footer h4


pokoli:
Thanks for contributing!

PR #580, opened 2018-10-08
Avoid use of node stream in browser context.

jefbarn:
Here I used uglify's conditional compilation to remove the Node stream code from the minified file.
Then reference the minified file under the 'browser' tag in package.json, which most front end packagers will pick up before 'main'.


jefbarn:
Also I couldn't run the phantomjs tests due to a bug between phantom and OSX.

papaparse.js
```diff
@@ -871,101 +871,104 @@ if (!Array.isArray)
 
 
 	function DuplexStreamStreamer(_config) {
```
pokoli: Does it really make sense to define a function with no code?

papaparse.js
```diff
@@ -871,101 +871,104 @@ if (!Array.isArray)
...
+			stream = new Duplex({
+				readableObjectMode: true,
+				decodeStrings: false,
+				read: bindFunction(this._onRead, this),
+				write: bindFunction(this._onWrite, this)
+			});
+			stream.once('finish', bindFunction(this._onWriteComplete, this));
+		}
 	}
 	DuplexStreamStreamer.prototype = Object.create(ChunkStreamer.prototype);
```
pokoli: Could we also include this part in the conditional definition?

papaparse.js
jefbarn: Yeah I don't really like it either, this was just a requirement of how the uglify conditional compilation works (can't have a function that's not at the top level).
I'll play around with it a little more and see if I can come up with something that makes more sense. Maybe just a console.warn that says that streams aren't supported on browser.

jefbarn:
Found a better solution to the conditionals that removes all reference to the function so dead code eliminator removes it.
Reverted the minified file, but not sure I understand the reasoning...the whole point of this fix is to change the minified file.

pokoli:
Not it looks good to me but I still see that the minified file is part of the PR, could you check please?

The reasoning to not include the minified file is the following:

- If you rebuild the minified file on your branch it does not include the new changes for other PR, so it also has to be rebuilded on each merge. In order to simplify the workflow we do not change it on each PR but only when uploading a new version to npm, so we ensure that all the right content is included on npm.

BTW: I plan to upload a new bugfix release on npm which includes it change once it's merged on master branch, so the minified file will be correctly generated.

arisk:
@jefbarn any update on removing the minified code from the PR? Thanks!

pokoli:
Merged manually in https://github.com/mholt/PapaParse/commit/0e392e2f2468aff246969a9102ff9762c37c2875 

And released [4.6.2](https://www.npmjs.com/package/papaparse/v/4.6.2) which includes this fix. 

Thanks for your contribution!

Issue #571, opened 2018-09-21
mocha-phantomjs is deprecated


MonkeyDZeke:
[Mocha-phantomjs is deprecated](https://github.com/nathanboktae/mocha-phantomjs) and it is recommended to upgrade to something like [mocha-chrome](https://github.com/shellscape/mocha-chrome).

I noticed this when I did an npm audit and found out that it's insecure. This is already documented [on the mocha-phantomjs repo](https://github.com/nathanboktae/mocha-phantomjs/issues/262).

pokoli:
Fixed by #581

PR #581, opened 2018-10-11
Switch to mocha-chrome

jacopofar:
As noticed in [this issue](https://github.com/mholt/PapaParse/issues/571) mocha-phantomjs [is deprecated](https://github.com/nathanboktae/mocha-phantomjs).

This PR replaces it with mocha-chrome, and updates mocha to the latest stable version.

Unfortunately, mocha-chrome [doesn't work on Node.js 6 and there are no plans to support it](https://github.com/shellscape/mocha-chrome/issues/10) since it uses `async`, and since the EOL of Node.js 6 is in 7 months I decided to use a fork of the project which is compatible with it as a test environment.


pokoli:
Hi @jacopofar, first of all thank you so much for creating this PR

I'm wondering if it's really worth to use a mocha-chrome fork to support older versions of Node.

The problem I see is that the fork is not updated with new mocha-chrome fixes which may produce unexpected errors that may be fixed in upstream. 

Thoughts?

pokoli:
I'm just created a discussion to include this on 5.0 release, so we can drop support for Node 6.0

jacopofar:
Yep, it's very annoying. I think there are three solutions:

1. Continue to use the deprecated phantomjs module
2. Drop test support for nodejs 6 and just use mocha-chrome
3. Fork the current version of `mocha-chrome` and remove the `async` references. There are not many so it should be feasible. I can take care of that.

I don't know how big is the usage of Node 6 nowadays, the only reference I have is the is the official EOL for it

jacopofar:
Ah OK, that could be a nice solution :)

pokoli:
@jacopofar could you update the PR to prevent using the fork and use the mocha-chrome relase?

I plan to merge it to master branch, which is the future 5.0 and which does not support node 6

jacopofar:
Done, but now a test apparently not related to the change fails :/

jsg2021:
Just fyi, `mocha-headless-chrome` has far less deps than `mocha-chrome`. :)

The failing test wasn't running on phantom (since it didn't support File) It now runs on Chrome. The test is only tracking the first index of each step's data... yet expecting the entire data... https://github.com/mholt/PapaParse/pull/599/commits/4877a0497e8464d6f8db0b5b6e10100c7dc44dd7#diff-c2c121651ee2e75f084ececb1c87516cL1961

pokoli:
I will prefer to use mocha-headless-chrome if it has less deps :)

jacopofar:
Switched, is now building. Except a new `-f` flag to pass the test page is basically a drop-in replacement :)

pokoli:
@jacopofar thanks for your work! I've merge it manually in 9a4a00a1ee8753eb2a106ef2de86d3a8a8bba698 and added an additional commit to fix the tests

PR #593, opened 2018-11-02
Modified the GuessDelimiter function.

Erik-HH:
Changed the GuessDelimiter function, so that a new delimiter is chosen as best if its delta value is higher than the current chosen.
Added a testcase for decimal values with comma as decimal separator and pipes as delimiter.
(Which should mimic the following test data:
a|3,4|b
c|3,4|d
)

But i couldnt ran the tests because i dont have a node interpreter installed, so i am not sure if the other tests break, altough the test data mentioned above now produces the correct result).

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -1262,7 +1262,7 @@ if (!Array.isArray)
 
 					if (typeof fieldCountPrevRow === 'undefined')
 					{
-						fieldCountPrevRow = fieldCount;
+						fieldCountPrevRow = 0;
 						continue;
 					}
 					else if (fieldCount > 1)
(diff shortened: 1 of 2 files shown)


pokoli:
Hi, thanks for the PR I see you've added a new test-cases.js file, you should modify the tests-cases.js file in the test folder.

pokoli:
We use tabs for formating and you just added some code with spaces. This [breaks the linter ](https://travis-ci.org/mholt/PapaParse/jobs/449809283), could you please fix it?

pokoli:
I will prefer to include only in this pull request the changes related to the GuessDelimiter function and move all the other code styling changes in a separate Pull Request.

PR #594, opened 2018-11-13
Add config.transformHeaders optional function

ihoro:
It can be used to rename headers found in CSV.
As an example, I use vue-tables-2 what treats dots in column naming as path to a property of a sub-object. And in my case third-party API responds with dots in columns naming. So, with this option I can easily replace dots with underscores and get everything working.


pokoli:
Thanks for your contribution but I'm closing this as a duplicate of #589

Issue #273, opened 2015-11-05
Worker does not work with a blob url


priyajeet:
Due to strict same origin restrictions for web workers the workaround to get it working is to make a GET request to the JS file and parse the response as a blob url.

``` javascript
fetch('http://some.cross.domain/path/to/papaparse.js')
.then((response) => response.blob())
.then((papaWorkerBlob) => {
      Papa.SCRIPT_PATH = URL.createObjectURL(papaWorkerBlob);
      Papa.parse(csvUrl, {
            worker: true,
            ...
      }
});
```

This can then be supplied to Papa.SCRIPT_PATH. However papaparse.js expects a query param `?papaworker` for the webworker and this does not seem to work well with blob urls.

Is there any way this can work? (other than duplicating the file and forcing one to be the worker)

priyajeet:
Another way I could get it working by changing the following line to not rely on `?papaparse`

```javascript
IS_PAPA_WORKER = IS_WORKER && typeof WorkerGlobalScope !== 'undefined' && self instanceof WorkerGlobalScope
````

And then obviously not appending `?papaparse`

pokoli:
Could you please provide a PR? Thanks in advance.

pokoli:
Fixed by #599

Issue #525, opened 2018-06-22
Failed to execute 'postMessage' on 'Worker'


dornstein:
I've been using papa parse successfully for quite a while.  If have found cases where it's necessary to parse really big (70MB) CSV file and while that's running the browser UI locks up.  So I thought I'd try 'worker: true'.  I added a complete: call back.  But when it runs I get the error: ***Failed to execute 'postMessage' on 'Worker'*** 

I'm running on Chrome 67.0.3396.87.

What am I missing?  The docs suggested all I needed was to say "worker: true"...

suavelizard:
This is happening in electron.

`
Error Parsing File
{
    "stack": "Error: Failed to execute 'postMessage' on 'Worker': function () {\n if (this.autoClose) {\n this.destroy();\n }\n } could not be cloned.\n at Object.CsvToJson [as parse] ([REMOVED]\\node_modules\\papaparse\\papaparse.js:225:6)\n at eval (webpack-internal:///./src/renderer/translators/shared.js:41:64)\n at new Promise ()\n at new F (webpack-internal:///./node_modules/core-js/library/modules/_export.js:36:28)\n at _callee$ (webpack-internal:///./src/renderer/translators/shared.js:40:20)\n at tryCatch (webpack-internal:///./node_modules/regenerator-runtime/runtime.js:62:40)\n at Generator.invoke [as _invoke] (webpack-internal:///./node_modules/regenerator-runtime/runtime.js:296:22)\n at Generator.prototype.(anonymous function) [as next] (webpack-internal:///./node_modules/regenerator-runtime/runtime.js:114:21)\n at step (webpack-internal:///./node_modules/babel-runtime/helpers/asyncToGenerator.js:17:30)\n at eval (webpack-internal:///./node_modules/babel-runtime/helpers/asyncToGenerator.js:35:14)"
}`

pokoli:
Fixed by #599

PR #599, opened 2018-11-19
[Rework Worker Definition] Make Bundling/Concatenation safe

jsg2021:
This addresses a few issues with using the worker and bundlers (such as rollup/webpack) I believe these issues are related and fixes: #273,  #426, #525

With this change, you could completely drop the SCRIPT_PATH option all together or default it to use the generated blob.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -62,8 +62,7 @@ if (!Array.isArray)
 	}
 
 	var IS_WORKER = !global.document && !!global.postMessage,
-		IS_PAPA_WORKER = IS_WORKER && (/blob:/i.test(global.location.protocol) || /(\?|&)papaworker(=|&|$)/.test(global.location.search)),
-		LOADED_SYNC = false, AUTO_SCRIPT_PATH;
+		IS_PAPA_WORKER = IS_WORKER && /blob:/i.test((global.location || {}).protocol);
 	var workers = {}, workerIdCounter = 0;
 
 	var Papa = {};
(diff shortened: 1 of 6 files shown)


papaparse.js
```diff
@@ -51,8 +54,19 @@ if (!Array.isArray)
 		return {};
 	})();
 
+
+	function getWorkerBlob() {
+		try {
```
pokoli: Why adding a try/catch? IIUC return null will make the code break on next line so it seems better to fail here directly in order to get a better stack trace

jsg2021: Yeah, I was on the fence about this. In practice it doesn't throw unless on older browsers, and in my practice I've nulled the value to indicate fallback behavior... but I can see the value in just blowing up here. Will update.

jsg2021:
Oh, if you are in favor of dropping the SCRIPT_PATH and papaworker=1 in this change, I can do that. It would simplify it greatly. :)

jsg2021:
I'm looking into the test failures.

jsg2021:
Hmm, the test 'Quoted line breaks near chunk boundaries are handled' wasn't running in phantom, but is running in other tests (and passing) Not sure why its failing in chrome.

pokoli:
I will prefer to resolve #581 first and then merge this PR when the other is resolved

pokoli:
@jsg2021 #581 is now on trunk (and test case is passing), so you should rebase your work on top of master in order to be able to merge it.

jsg2021:
Rebased :)

pokoli:
Thanks for your work. I've merged it manually on https://github.com/mholt/PapaParse/commit/4ecae0b5bbcc76a16abecded59792ed9cea750cb because I prefere to leave 296c89049bcb5f6ab0dc7008b326d641e1d6a531 as separate commit (as it's not related).

jsg2021:
This replaces #428 and resolves #273, #426, #525 all of which should be able to be closed now :)

PR #621, opened 2019-01-25
Fixes #619

### LINKED ISSUE #619: 5.0.0-beta.0 header: true, worker: false, steps... incorrectly parses headers
Here is a failing test:
```
AssertionError: expected '[[{"O":"A"},{"O":"b"},{"O":"c"}],[{"O":"d"},{"O":"E"},{"O":"f"}]]' to deeply equal '[{"One":"A","Two":"b","Three":"c"},{"One":"d","Two":"E","Three":"f"}]'
      + expected - actual

      -[[{"O":"A"},{"O":"b"},{"O":"c"}],[{"O":"d"},{"O":"E"},{"O":"f"}]]
      +[{"One":"A","Two":"b","Three":"c"},{"One":"d","Two":"E","Three":"f"}]
```
<details>
<summary>code</summary>

```js
	{
		description: "Data is correctly parsed with steps (headers)",
		expected: [{One: 'A', Two: 'b', Three: 'c'}, {One: 'd', Two: 'E', Three: 'f'}],
		run: function(callback) {
			var data = [];
			Papa.parse('One,Two,Three\nA,b,c\nd,E,f', {
				header: true,
				step: function(results) {
					data.push(results.data);
				},
				complete: function() {
					callback(data);
				}
			});
		}
	},
	{
		// The worker code path does work,
		// This fails because node doesn't expose Workers yet ... so this falls back to the non-worker path.
		// See: https://nodejs.org/api/worker_threads.html
		description: "Data is correctly parsed with steps and worker (headers)",
		expected: [{One: 'A', Two: 'b', Three: 'c'}, {One: 'd', Two: 'E', Three: 'f'}],
		run: function(callback) {
			var data = [];
			Papa.parse('One,Two,Three\nA,b,c\nd,E,f', {
				header: true,
				worker: true,
				step: function(results) {
					data.push(results.data);
				},
				complete: function() {
					callback(data);
				}
			});
		}
	},
```

</details>

jsg2021:
Adds tests and fixes an issue that incorrectly parses the headers with step functions

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -1125,18 +1125,25 @@ License: MIT
 		{
 			if (!_results)
 				return;
-			for (var i = 0; needsHeaderRow() && i < _results.data.length; i++)
-				for (var j = 0; j < _results.data[i].length; j++)
-				{
-					var header = _results.data[i][j];
 
-					if (isFunction(_config.transformHeader)) {
-						header = _config.transformHeader(header);
-					}
+			function addHeder(header)
+			{
+				if (isFunction(_config.transformHeader))
+					header = _config.transformHeader(header);
 
-					_fields.push(header);
-				}
-			_results.data.splice(0, 1);
+				_fields.push(header);
+			}
+
+			if (Array.isArray(_results.data[0]))
+			{
+				for (var i = 0; needsHeaderRow() && i < _results.data.length; i++)
+					_results.data[i].forEach(addHeder);
+
+				_results.data.splice(0, 1);
+			}
+			// if _results.data[0] is not an array, we are in a step where _results.data is the row.
+			else
+				_results.data.forEach(addHeder);
 		}
 
 		function shouldApplyDynamicTyping(field) {
@@ -1170,15 +1177,15 @@ License: MIT
 			if (!_results || (!_config.header && !_config.dynamicTyping && !_config.transform))
 				return _results;
 
-			for (var i = 0; i < _results.data.length; i++)
+			function processRow(rowSource, i)
 			{
 				var row = _config.header ? {} : [];
 
 				var j;
-				for (j = 0; j < _results.data[i].length; j++)
+				for (j = 0; j < rowSource.length; j++)
 				{
 					var field = j;
-					var value = _results.data[i][j];
+					var value = rowSource[j];
 
 					if (_config.header)
 						field = j >= _fields.length ? '__parsed_extra' : _fields[j];
@@ -1197,7 +1204,6 @@ License: MIT
 						row[field] = value;
 				}
 
-				_results.data[i] = row;
 
 				if (_config.header)
 				{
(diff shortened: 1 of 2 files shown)


papaparse.js
```diff
@@ -1111,32 +1110,38 @@ License: MIT
...
+				if (isFunction(_config.transformHeader))
+					header = _config.transformHeader(header);
 
-					_fields.push(header);
-				}
-			_results.data.splice(0, 1);
+				_fields.push(header);
+			}
+
+			if (isStep)
```
pokoli: can we test that _config.step is set instead of passing the isStep variable everywhere?

papaparse.js
```diff
@@ -1165,20 +1170,19 @@ License: MIT
...
 			{
 				var row = _config.header ? {} : [];
 
-				var j;
-				for (j = 0; j < _results.data[i].length; j++)
+				function processField(value, j)
 				{
 					var field = j;
-					var value = _results.data[i][j];
+					// var value = _results.data[i][j];
```
pokoli: Why comenting?

papaparse.js
```diff
@@ -1165,20 +1170,19 @@ License: MIT
...
 		}
 
-		function applyHeaderAndDynamicTypingAndTransformation()
+		function applyHeaderAndDynamicTypingAndTransformation(isStep)
 		{
 			if (!_results || (!_config.header && !_config.dynamicTyping && !_config.transform))
 				return _results;
 
-			for (var i = 0; i < _results.data.length; i++)
+			function processRow(rowSource, i)
```
pokoli: Could you avoid refactoring the code into function instead of fors?

I do not say i don't like it but i like to keep commits simple and related to the issue. 

Please open a new PR for this kind of refactoring.

papaparse.js
```diff
@@ -1165,20 +1170,19 @@ License: MIT
...
 			{
 				var row = _config.header ? {} : [];
 
-				var j;
-				for (j = 0; j < _results.data[i].length; j++)
+				function processField(value, j)
```
pokoli: idem about refactoring

tests/test-cases.js
```diff
@@ -1804,6 +1804,42 @@ var CUSTOM_TESTS = [
...
+					data.push(results.data);
+				},
+				complete: function() {
+					callback(data);
+				}
+			});
+		}
+	},
+	{
+		// The worker code path does work,
```
pokoli: I do not understand what is this comment talking about. 

The tests are executed under mocha-chorme and I see this [is psasing on travis.](https://travis-ci.org/mholt/PapaParse/jobs/484479176#L901)

jsg2021: That comment was made while the test was failing... and I didn't realize there was a chrome headless test ...i will remove it.

papaparse.js
jsg2021: I was not intending to refactor for refactors sake. If the loop body is not put into a function, it will need to be repeated. I attempted to make the change as minimal as possible.

papaparse.js
jsg2021: again, this wasn't a refactor as much as it was reusing the loop body

papaparse.js
jsg2021: sorry, that was sloppy. I meant to remove it.

papaparse.js
jsg2021: will do.

jsg2021: Okay. this is a problem. `_config.step` is truthy even when you are processing outside of a step...  the problem we are attempting to fix is that the step function calls on the process/parse functions that assume the shape of results is a two dimensional array, when for steps, it's just a single dimension.  We need to tell the those functions the context is one row. not a chunk of rows... Alternatively we could ensure steps use a two dimensional array as well, but limit the length to 1 row... but that is way more invasive at the moment.

pokoli: Then probably is better to make those functions always work with a one dimensional array.

jsg2021: So, are you suggesting to test the shape of `_results.data` instead of passing in the flag?

jsg2021: done. :)

jsg2021:
I think I've addressed your requests? If not let me know :)

PR #628, opened 2019-02-07
add table for unparse config options

janisdd:
This adds a dedicated section/block for the unparse config option.
I think this way it is more similar to the parse config.

It also clarifies some unparse options (e.g. `quotes` can be an array of booleans).


pokoli:
Thank you so much!!!!

PR #629, opened 2019-02-07
Add mailcheck.co to lovers.js

nosovk:
Papa Parser is a great help for us. Such a function like guessing delimiters saved lots of our clients time.


mholt:
Thank you! Glad you find it useful.

Issue #609, opened 2018-12-20
Unparse skips columns


silvestreh:
Looks like unparsing a JSON array ends up in missing columns if the first `Object` in the array is missing properties that other `Object`s in the same `Array` do have.

This `Array` is correctly turned into CSV:

```javascript
[
  {
    date: '2018-12-20T17:04:41.446Z',
    amount: '30.03',
    metadata_customerEmail: 'someone@domain.com'
  },
  {
    date: '2018-12-20T17:01:30.434Z',
    amount: '13.06',
    metadata_customerEmail: 'someone@domain.com'
  },
  {
    date: '2018-12-20T16:49:55.630Z',
    amount: '31.33',
    metadata_customerEmail: 'someone@domain.com'
  },
  {
    date: '2018-12-20T16:33:50.121Z',
    amount: '29'
  }
]
```

This is the output:

```csv
date,amount,metadata_customerEmail
2018-12-20T17:04:41.446Z,30.03,someone@domain.com
2018-12-20T17:01:30.434Z,13.06,someone@domain.com
2018-12-20T16:49:55.630Z,31.33,someone@domain.com
2018-12-20T16:33:50.121Z,29,
```

Now, if you `sort` that array so the oldest entry (the one without `metadata_customerEmail`) is at `index` 0, then you end up with this CSV:

```csv
date,amount
2018-12-20T16:33:50.121Z,29
2018-12-20T16:49:55.630Z,31.33
2018-12-20T17:01:30.434Z,13.06
2018-12-20T17:04:41.446Z,30.03
```

silvestreh:
Here's a thought: maybe Papa could make a first pass to determine what the columns should be and then a second one to grab all the data?

pokoli:
I don't think we should double pass the results as this will be a performance issue but you may probably do it before passing the values to paparse. This way you can perform any additional checks or whatever you need. 

Do you have some suggestion about how to improve the unparse of empty values? For me the current behavior is correct.

MonkeyDZeke:
I would suggest we add a config variable for unparse called something like `columns` which accepts either an array (`['key_1', 'key_2', 'key_3']`) or an object (`{old_key_1: 'new_key_1', old_key_2: 'new_key_2', old_key_3: 'new_key_3'}`) so that the user can tell the unparser what to expect (and conveniently alter the resulting header row if desired).

Thoughts?

dboskovic:
[...]

PR #632, opened 2019-02-09
Fixes issue #609

janisdd:
Introduces a new `unparse` option `columns` to specify the columns/keys manually.

Some points i'm not sure about:

- Default value is currently null 
I think `[]` is not a good choice because then we need to check `_columns.length > 0` in order to know if the user specified a custom value. If someone des a `array.map()` for the keys that results into `[]` then the keys of the first object are used (which is unexpected).
However, maybe `null` looks a bit strange as default value

- I'm not sure about the last test... is the expected value correct?

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -291,10 +291,7 @@ License: MIT
 			if (!_input.length || Array.isArray(_input[0]))
 				return serialize(null, _input, _skipEmptyLines);
 			else if (typeof _input[0] === 'object')
-				return serialize(
-					_columns ? _columns : objectKeys(_input[0]),
-					_input, _skipEmptyLines
-				);
+				return serialize(_columns || objectKeys(_input[0]), _input, _skipEmptyLines);
 		}
 		else if (typeof _input === 'object')
 		{
@@ -351,6 +348,9 @@ License: MIT
 				_writeHeader = _config.header;
 
 			if (Array.isArray(_config.columns)) {
+
+				if (_config.columns.length === 0) throw new Error('Option columns was empty');
+
 				_columns = _config.columns;
 			}
 		}
--- a/papaparse.js
+++ b/papaparse.js
@@ -349,7 +349,7 @@ License: MIT
 
 			if (Array.isArray(_config.columns)) {
 
-				if (_config.columns.length === 0) throw new Error('Option columns was empty');
+				if (_config.columns.length === 0) throw new Error('Option columns is empty');
 
 				_columns = _config.columns;
 			}
--- a/papaparse.js
+++ b/papaparse.js
@@ -276,6 +276,9 @@ License: MIT
 		/** whether to skip empty lines */
 		var _skipEmptyLines = false;
 
+		/** the columns (keys) we expect when we unparse objects */
+		var _columns = null;
+
 		unpackConfig();
 
 		var quoteCharRegex = new RegExp(escapeRegExp(_quoteChar), 'g');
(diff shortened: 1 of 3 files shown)


papaparse.js
```diff
@@ -288,7 +291,10 @@ License: MIT
 			if (!_input.length || Array.isArray(_input[0]))
 				return serialize(null, _input, _skipEmptyLines);
 			else if (typeof _input[0] === 'object')
-				return serialize(objectKeys(_input[0]), _input, _skipEmptyLines);
+				return serialize(
+					_columns ? _columns : objectKeys(_input[0]),
```
pokoli: I think we can use _columns || objectKeys(input[0]) which will remove the ? and we are able to set all arguments in the same line

tests/test-cases.js
```diff
@@ -1732,6 +1732,20 @@ var UNPARSE_TESTS = [
...
+		notes: "Should not throw any error when attempting to serialize key not present in object. Columns are different than keys of the first object. When an object is missing a key then the serialized value should be an empty string.",
+		input: [{a: 1, b: '2'}, {}, {a: 3, d: 'd', c: 4,}],
+		config: {columns: ['a', 'b', 'c']},
+		expected: 'a,b,c\r\n1,2,\r\n\r\n3,,4'
+	},
+	{
+		description: "Column option is empty",
+		notes: "Header (columns) is empty so header is skipped",
+		input: [{a: 1, b: '2'}, {}, {a: 3, d: 'd', c: 4,}],
+		config: {columns: []},
```
pokoli: Does it really make sense to specify an empty array in columns? It's not better to ignore the columns setting in this case and return the keys of first object?

janisdd: Hm if you pass in `[]` by accident and get good looking data back, one would think everything works fine (even if the array was `[]` and we defaulted to the keys of the first object). I think it would help to discover bugs...but not bothering the user is also a good idea...

pokoli: I see your rational but using [] means not specifying any column so no file can be generated so it's a nonesense, so for me the parameter should be ignored or an error raised

papaparse.js
```diff
@@ -343,6 +346,13 @@ License: MIT
 
 			if (typeof _config.header === 'boolean')
 				_writeHeader = _config.header;
+
+			if (Array.isArray(_config.columns)) {
+
+				if (_config.columns.length === 0) throw new Error('Option columns was empty');
```
pokoli: was -> is

tests/test-cases.js
```diff
@@ -1740,6 +1754,14 @@ describe('Unparse Tests', function() {
 		(test.disabled ? it.skip : it)(test.description, function() {
 			var actual;
 
+			if (test.expectsError) {
```
pokoli: I do not think is required to test the error.

tests/test-cases.js
```diff
@@ -1732,6 +1732,20 @@ var UNPARSE_TESTS = [
...
+	},
+	{
+		description: "Column option used to manually specify keys",
+		notes: "Should not throw any error when attempting to serialize key not present in object. Columns are different than keys of the first object. When an object is missing a key then the serialized value should be an empty string.",
+		input: [{a: 1, b: '2'}, {}, {a: 3, d: 'd', c: 4,}],
+		config: {columns: ['a', 'b', 'c']},
+		expected: 'a,b,c\r\n1,2,\r\n\r\n3,,4'
+	},
+	{
+		description: "Column option is empty",
```
pokoli: For me we can remove this test.

janisdd:
Ok, done.

PR #650, opened 2019-03-28
support custom decoder option.

grizim:
used for cases when the encoding isn't supported by native buffer,  for example parsing 'SJIS'  in node.

tested by parsing fileReadSream with encoding SJIS with chunks on Node v10.13.0:
```
Papa.parse(fileReadStream, {
            encoding: encoding,
            delimiter: delimiter,
            decoder: function(buf , encoding) {
                if (Buffer.isEncoding(encoding)) {
                    return buf.toString(encoding)
                }
                if (iconv.encodingExists(encoding)) {
                    return iconv.decode(buf, encoding)
                }
                throw new TypeError('[ERR_UNKNOWN_ENCODING]', 'Unknown encoding: ' + encoding)
            },
            complete: () => {},
            chunk: function (chunk) {
                console.log(JSON.stringify(chunk.data))
            }
        })
```


pokoli:
Thanks for your pull request!

Could you please add some test case that tests this behaviour?

grizim:
Hi @pokoli, 
I added some tests but I'm not sure they are enough.

My tests use "this._streamData" method , but not "_addToParseQueue". 
I can see that "_addToParseQueue" tested by "piped streaming CSV should be correctly parsed" but the chunk is a string and not a buffer so there is no need to use the "decoder".

package.json
```diff
@@ -42,6 +42,7 @@
 		"eslint": "^4.19.1",
 		"grunt": "^1.0.2",
 		"grunt-contrib-uglify": "^3.3.0",
+		"iconv-lite": "^0.4.24",
```
pokoli: Can we avoid adding a new dependency? As it's only used on test maybe we can use a simpler test

docs/docs.html
```diff
@@ -530,6 +531,20 @@ <h5 id="config-details">Config Options</h5>
 									The encoding to use when opening local files. If specified, it must be a value supported by the FileReader API.
 								</td>
 							</tr>
+							<tr>
+								<td>
+									<code>decoder</code>
+								</td>
+								<td>
+									A function which allows custom decoding of the chunk.
```
pokoli: Could we explain what's the usage of this function. For me it is not clear when I should set the decoder option.

docs/docs.html
```diff
@@ -530,6 +531,20 @@ <h5 id="config-details">Config Options</h5>
...
+									The function get Chunk (Buffer/string) and Encoding, and returns a string.
+									i.e assuming const iconv = require('iconv-lite'):
```
pokoli: I do not think it's needed to add an example but better to explain the usage.

papaparse.js
```diff
@@ -852,7 +852,18 @@ License: MIT
...
 			try
 			{
-				queue.push(typeof chunk === 'string' ? chunk : chunk.toString(this._config.encoding));
+				var encoding = this._config.encoding;
+				var decoder = this._config.decoder;
+				var decodedString = '';
+
+				if (typeof decoder === 'function') {
+					decodedString = decoder(chunk, encoding);
+				} else if (typeof chunk === 'string') {
```
pokoli: For me it is strange to use an if/else with different variables. 

What if decoder is not a function?

papaparse.js
```diff
@@ -941,7 +952,19 @@ License: MIT
 			// when too many items have been added without their
 			// callback being invoked
 			parseCallbackQueue.push(bindFunction(function() {
-				this.parseChunk(typeof chunk === 'string' ? chunk : chunk.toString(config.encoding));
+				var encoding = config.encoding;
```
pokoli: I see this code is an exact copy of the other version. 
It's not better to encapsulate it on a function?

tests/node-tests.js
```diff
@@ -272,4 +272,89 @@ describe('PapaParse', function() {
...
+				if (isFirstLine) {
+					isFirstLine = false;
+					assert.notDeepEqual(results.data, [["コントロールカラム","商品管理番号（商品URL）"]]);
+				} else {
+					done();
+				}
+			},
+			encoding: 'utf-8',
+			decoder: function(buffer, encoding) {
+				if (iconv.encodingExists(encoding)) {
```
pokoli: Does it makes sense to use this function as default value for decoding? 
Which other functions can be used?

grizim:
Hi, I see that I can solve my problem without this pull-request by decoding the original read stream.
instead of:
```
Papa.parse(fs.createReadStream(__dirname + '/sjis-sample.csv')), { encoding 'sjis', ... }
```
pipe the original stream with iconv.decodeStream:
```
const iconv = require('iconv-lite')
Papa.parse(fs.createReadStream(__dirname + '/sjis-sample.csv').pipe(iconv.decodeStream('sjis')), { ... }
```

I did a small test and it looks fine, later I'll check it with a real use case and cancel the pull-request if it is fine.

docs/docs.html
grizim: Nodejs's Buffer support only a few encodings ('ascii','utf8','utf16le','ucs2','base64','latin1','binary','hex') while in the browser the fileReader supporting much more (https://encoding.spec.whatwg.org/) . 

when I tried to parse a stream with Papaparse I got "[ERR_UNKNOWN_ENCODING] Unknown encoding: SJIS" error because Buffer isn't supporting 'sjis' encoding.

grizim:
@pokoli , looks that decoding the stream before papaparse with iconv-lite solved my problem. 
do you want me to fix the code? or just cancel the pull request?

pokoli:
As it's something that it can be achieved after papararse I do not think we should include it the library. 
One of our main objectives is to no increase the complexity of the library.

I'm closing this PR, thanks for your time.

Issue #655, opened 2019-04-19
"window is not defined" error when worker is set to true


jjech:
This is a duplicate of #585.  When worker is set to true, main.chunk.js?papaworker throws an uncaught reference error: "window is not defined."

If this is a configuration issue, more information is needed about how to resolve it. In #585 @pokoli referenced [Can Papa Parse be loaded asynchronously (after the page loads)? section](https://www.papaparse.com/faq#combine) of the FAQ, but the FAQ in its current state doesn't mention anything about the PAPA_SCRIPT path variable.

This appears to be a common problem when workers are enabled in PapaParse, as several users have been looking for help in #585.

![Capture](https://user-images.githubusercontent.com/3535684/56436566-ed2e4500-62a1-11e9-9629-3a0f55ac35f3.PNG)

jjech:
Related issues: #563 #566 

When I set the SCRIPT_PATH manually, I get the following error:
papaparse?papaworker:1 Uncaught SyntaxError: Unexpected token <

Someone else experienced this issue, too...
https://stackoverflow.com/questions/53782542/papaparse-script-path-is-undefined

I think maybe we need to "use webpack to serve papaparsejs as static file, so do not wrap it with webpack module" as @tamnvt mentions in #566 but I don't quite understand this comment.

jjech:
Got it... here are the final steps that worked for me.
1. Copied papaparse.js from node_modules/papaparse/ to my main app directory.
2. Added the following lines to both App.js and index.js

var Papa = require("papaparse");
Papa.SCRIPT_PATH = "./papaparse.js";

jjech:
Actually, I'm going to reopen this as I think it should be documented on the FAQ. Feel free to close if you disagree...

pokoli:
@jjech it will be great if you add a PR documenting this behaviour. 
This faq source code is available on the[ faq.html file on the docs folder](https://github.com/mholt/PapaParse/blob/master/docs/faq.html).

jjech:
Merged PR #656, closing issue.

PR #656, opened 2019-04-23
Update FAQ with reference to configuration issue #655

jjech:
Since several users have reported an error when setting worker:true, update FAQ with reference to issue #655.


pokoli:
Thank you so much!

Issue #658, opened 2019-04-26
Delimiter not auto-detected correctly when using OPTIONAL quoted fields.


konuch:
When trying to auto-detect the delimiter from following line:  `Field1;Field2;"Field,3,3,3"`, it wrongly guesses the delimiter as COMMA instead of SEMICOLON, even if the field with comma is quoted.
However, it does work correctly, when all 3 fields are quoted:  `"Field1";"Field2";"Field,3,3,3"`.
What is strange, it also works correctly when only 1st and 3rd field is quoted: `"Field1";Field2;"Field,3,3,3"`.
Tested on demo site: https://www.papaparse.com/demo

Expected behavior would be that the guessing logic correctly takes a quoted string as a field and would not consider its characters when guessing the delimiter.

The description for `quoteChar` option states: `... Any field which is not quoted will correctly read.` This makes the user believe, that he does not need to enclose all fields in quotes, in order to make this use case work.

PS. Enclosing all fields to quotes is not an option in our business use case.

pokoli:
Any PR that improves the auto-detect feature will be very wellcome. Just make sure to include a test case to ensure everything is working.

pokoli:
Fixed in #665.

PR #665, opened 2019-05-27
Handle delimiter guessing, when not all of the fields are quoted

konuch:
PR for https://github.com/mholt/PapaParse/issues/658


pokoli:
Thank you so much!

PR #669, opened 2019-06-01
fix  #668, update webiste to v5

### LINKED ISSUE #668: Documentation updated for v5?
Thanks for keeping PP updated!

[Website still says  v4](https://github.com/mholt/PapaParse/blob/master/docs/docs.html#L37) though. ... Sorry, I don't know PP enough to see if there are changes needed (aside from version bump).

tomByrer:
Might want to [check the rest of the docs](https://github.com/mholt/PapaParse/issues/668); I only changed 1 character ;)

But looking at the file history, seems you kept it up.

### DIFF
--- a/docs/docs.html
+++ b/docs/docs.html
@@ -34,7 +34,7 @@
 						</div>
 					</div>
 					<div class="grid-20 hide-on-mobile text-center">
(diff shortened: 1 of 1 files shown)


pokoli:
Thank you so much

PR #678, opened 2019-06-13
Fix invalid link to lovers.js

hughanderson4:
https://www.papaparse.com/ currently has 404 link to lovers.js, let's fix it.


pokoli:
Thank you so much!

PR #679, opened 2019-06-17
Make the demo website use the new 5.0.0 version.

konuch:


pokoli:
I'm wondering if it won't be better to load the file directly from github master branch. 

What do you think?

Issue #675, opened 2019-06-07
Issues in new version of papaparse


kroupacz:
Hello,
after upgrade to paparse 5.0.0, I have a problem with correct delimiter detection and "header": true config. It look like this issue is related to this commit/change: https://github.com/mholt/PapaParse/commit/f6b1b36c3f8faf5ab8e436709925e1563b0f75de

Below I am sending two files where the problem occurs.
- http://drive.google.com/uc?export=download&id=1NjL0RQ64uD6gpaLfrnSryLuLAzeu-Hm9 - invalid delimiter detection

- http://drive.google.com/uc?export=download&id=1ZBSSBIROPt2w1PsqrWmvgImFBCBxzntr - Incorrect parsing with "header": true config

Both of these files are correctly parsed in papaparse version 4.6.1.

Thank you for fixing this issue.
Tomáš

georgeslabreche:
I believe this is the same issue: https://github.com/mholt/PapaParse/issues/685

pokoli:
Could you test this is fixed on current master? #687 may have fixed this.

seth-shaw-unlv:
I don't believe #685 is a duplicate of this issue. I've posted my example on that thread and the current master doesn't resolve the issue.

georgeslabreche:
I do apologize for mistakingly relating the two different issues!

martinstreicher:
I have a perfectly valid CSV file that will not upload properly in Papa 5 with `header: true`. It worked fine previously. However, if I also enable the `Skip empty lines` feature, the file processes just fine. Oddly though, there are no blank lines in the file. Further, the error in parsing always seems to occur on the final line of the file.

pokoli:
@martinstreicher It will be great if you can share the file or add a simple test case to reproduce the issue. Otherwise it is not possible to provide a fix for it.

PR #687, opened 2019-06-27
Modified the GuessDelimiter function

morance:
question: function return different cause the version change

input:"a,b,c
a,b,c|d|e|f"

output:[["a", "b", "c"], ["a", "b", "c|d|e|f"]] in v4
              [["a,b,c"],["a,b,c","d","e","f",]] in v5

expect :[["a", "b", "c"], ["a", "b", "c|d|e|f"]]

resolve:  delta should  record the difference between line which mean smaller is better, when delta is equal, we should compare the avgfieldcount and choose the biggest one.by the way the issue592 can resolve right array use this function.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -1261,14 +1261,12 @@ License: MIT
 			return _results;
 		}
 
-		function guessDelimiter(input, newline, skipEmptyLines, comments, delimitersToGuess)
-		{
-			var bestDelim, bestDelta, fieldCountPrevRow;
+		function guessDelimiter(input, newline, skipEmptyLines, comments, delimitersToGuess) {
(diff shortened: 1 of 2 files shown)


pokoli:
Hi @morance, 

First of all thank you so much for this pull request. 

Could you please add the example on your description to [our test suite](https://github.com/mholt/PapaParse/blob/master/tests/test-cases.js). 
So this way we ensure this behaviour stays consistent in the next versions. 

Once added I will be very happy to merge it.

pokoli:
I think this will solve #675

morance:
ok, I add the test suite already. hope you can adopt it Thanks

pokoli:
Thank you so much!

PR #693, opened 2019-07-21
Update Flatfile with latest PapaParse master

dboskovic:


dboskovic:
Ooops, wrong direction of PR

Issue #692, opened 2019-07-16
Dynamic Typing changes bit int fields


lebedevat:
For `dynamicTyping: true` parser changes big int fields. For example, file below without dynamicTyping
```
["208310224003697457", "1", "0"]
["208310224003697685", "1", "0"]
```
with dynamicTyping
```
[208310224003697470, 1, 0]
[208310224003697700, 1, 0]
```
[big_int_example.txt](https://github.com/mholt/PapaParse/files/3394747/big_int_example.txt)

pokoli:
I think big integer fields are not supported in dynamic typing, so a pull request adding support for it will very welcome.

haxxxton:
literally just came across this same problem, will investigate a pull request next week if its not already done

pokoli:
Fixed in #694

PR #694, opened 2019-07-26
Leave big numbers as strings (>2^53 || <-2^53)

haxxxton:
Issues like #692 occur when numbers pass the regex for being numerical, but then are too large to have their precision maintained when using `parseFloat`. To avoid needing to use a library like Big to handle numbers of this size (IMO its a lot of overhead) we instead leave them as strings, and instead would encourage users to use a custom dynamicTyping function if they need to handle them as numbers.

### DIFF
--- a/docs/docs.html
+++ b/docs/docs.html
@@ -508,7 +508,7 @@ var csv = Papa.unparse({
 									<code>dynamicTyping</code>
 								</td>
 								<td>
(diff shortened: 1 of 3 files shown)


pokoli:
Thank you so much for your contribution!

haxxxton:
@pokoli do you have an ETA for when the next release is likely to go out? would love to use this in production from the offical repo rather than my fork :)

pokoli:
@haxxxton there is no ETA for the next release. Usually we wait a little bit to see if new features come to the PR.

You can use our master branch meanwhile

Issue #685, opened 2019-06-24
parsing gets wrong if first line begins with a comma


clombardi:
Just check the following:
```
papa.parse(',"a","b"\n,,"c","d,e",,34,')
```
The delimiter is the comma, this is OK. But the parsed first line includes just one column that encompasses the whole string `',"a","b"'` 

The problem seems to happen with the first line only, leading commas in subsequent lines make no hassle.

In the program where I use papaparse, I fixed this problem by adding a leading `""` if the string starts with a comma, hence I get
```
papa.parse('"","a","b"\n,,"c","d,e",,34,')
```
which works OK

georgeslabreche:
I am also encountering this issue. I am working around it by setting the header option to false, manually skipping the header row, and accessing data with array indices.

pokoli:
OK, so i close this issue as duplicate of #675

clombardi:
... if you consider that the two issues refer to the same problem, it is
OK, whenever you fix it in some future version. I had checked #675, it was
not clear to me whether it raised the same issue that happened to me. I
take the opportunity to congratulate the Papaparse team, is a very nice
tool. Cordially - Carlos Lombardi

seth-shaw-unlv:
I don't believe this is the same issue as #675. I have also run into this issue where, when the first value of the first row is blank, it fails to parse.

For example, the spreadsheet
```
,myfile.jpg,My nice image,2019-07-01?,"A fine image, yes?",Do whatever you want with it.,There's only one image.,29;28,Anonymous,,
,anotherFile.tif,My Kool Picture,2019-07-02,A kool picture.,Public Domain,55 mb,27;28,,,
,,,,,,,,,,
```
does not parse correctly:
![Screen Shot 2019-07-05 at 4 43 56 PM](https://user-images.githubusercontent.com/29869988/60748730-d9f85080-9f45-11e9-98a8-9bdeb3d2eb1f.png)

but changing the first value of the first row to a non-blank value (e.g. "b") causes the parser to work correctly:
![Screen Shot 2019-07-05 at 4 41 05 PM](https://user-images.githubusercontent.com/29869988/60748744-feecc380-9f45-11e9-9a0e-9ca3906cf071.png)

I've tested this with the [current master version](https://github.com/mholt/PapaParse/blob/master/papaparse.js) as [mentioned in the other issue thread](https://github.com/mholt/PapaParse/issues/675#issuecomment-506947859) but the error persists.

[...]

PR #696, opened 2019-07-28
Bugfix: Resolve parsing issue when first field is empty and unquoted

dboskovic:
This resolves the issue mentioned in #685

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -1606,6 +1606,7 @@ License: MIT
 					if (quoteSearch !== -1) {
 						// we have quotes, so we try to find the next delimiter not enclosed in quotes and also next starting quote char
 						var nextDelimObj = getNextUnqotedDelimiter(nextDelim, quoteSearch, nextNewline);
+
 						// if we have next delimiter char which is not enclosed in quotes
 						if (nextDelimObj && typeof nextDelimObj.nextDelim !== 'undefined') {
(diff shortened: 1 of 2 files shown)


thomasjulianstoelen:
@dboskovic hey david just curious, how come we continue on as an empty string instead of throwing an error saying a header is missing?

dboskovic:
There’s two layers here. One where we parse which should absolutely allow for anything in the RFC and be forgiving beyond. In this case empty data should never be a problem.

I think when you enable header mapping, we could introduce a strict mode that produces errors in the case of an empty header (or multiple identical headers) that could cause unexpected data in the end result. That said, I don’t think it’s unreasonable for there to be unlabeled columns in a dataset, so My gut says that type of validation would be better living downstream from this library.

pokoli:
@dboskovic Thanks for the patch!

ynicolaou:
@dboskovic this fix doesn't seem to work when using a step function. I noticed a few things that don't get parsed properly when using the step function. Any idea why is that?

pokoli:
@ynicolaou could you plese open a new issue with some example in order to reproduce it?

ynicolaou:
@pokoli see https://github.com/mholt/PapaParse/issues/697

grizim:
@pokoli  - can you please publish it to NPM?

pokoli:
I've published 5.0.2 which contains the fix.

PR #703, opened 2019-08-26
Implemented quotes config optionally as function

Puzzleton:
Fix for issue #532. In the thread for #534, it was suggested that the `quotes` parameter be implemented as a function.


tests/test-cases.js
```diff
@@ -1711,6 +1711,30 @@ var UNPARSE_TESTS = [
...
+		expected: '"a","b","c"\r\n"d",10,true'
+	},
+	{
+		description: "Force quotes around string fields only (with header row)",
+		input: [{ "Col1": "a", "Col2": "b", "Col3": "c" }, { "Col1": "d", "Col2": 10, "Col3": true }],
+		config: { quotes: function(value) { return typeof value === 'string'; } },
+		expected: '"Col1","Col2","Col3"\r\n"a","b","c"\r\n"d",10,true'
+	},
+	{
+		description: "Force quotes around values in first column only",
```
pokoli: I think those test are redundant as they test the same functionality but using a diferent function. 

Could you please remove them and include only the first two. 

Limiting the number of required test functions reduces the maintentance cost when refactoring the aplication.

docs/docs.html
```diff
@@ -282,7 +282,7 @@ <h5>Unparse Config Options</h5>
 									<code>quotes</code>
 								</td>
 								<td>
-									If <code>true</code>, forces all fields to be enclosed in quotes. If an array of <code>true/false</code> values, specifies which fields should be force-quoted (first boolean is for the first column, second boolean for the second column, ...).
+									If <code>true</code>, forces all fields to be enclosed in quotes. If an array of <code>true/false</code> values, specifies which fields should be force-quoted (first boolean is for the first column, second boolean for the second column, ...). If a function (accepting cell value & column index as parameters), returning <code>true</code> will force quotes for that value and <code>false</code> will not.
```
pokoli: What about: 

> A function that returns a boolean values can be used to determine the quotes value of a cell. This function accepts the cell value and column index as parameters.

tests/test-cases.js
pokoli: They are redundant because the tested functionality is exactly the same, what only changes is the quotes function which is an user parameter which we do not care about it. Our code should work for all the user provided functions. 

So please, include only the first two test.

pokoli:
Thanks for your contribution!

PR #709, opened 2019-09-17
Fix Range Header processing logic for NetworkStreamer

t83714:
Not all HTTP servers implement the correct behaviour of processing Range header for partial download.

Some HTTP server may ignore the range-end parameter and return the file content from the start index to the end of the file.

e.g.:
Request:
```
Range: bytes=0-2097151
Sec-Fetch-Mode: cors
User-Agent: Mozilla/5.0 
```
Response:
```
accept-ranges: bytes
Content-Length: 8256106
Content-Range: bytes 0-8256106/8256106
server: nginx/1.13.12
status: 206, 206 Partial Content
```

As we increase the `_start` index by `chunkSize` config option at each round of chunk processing (instead of increasing the `_start` by the actual size of the response), papaparse may incorrect re-download & parse unnecessary chunk:

https://github.com/mholt/PapaParse/blob/94d7bf939a210063d42551179ad44094b3013684/papaparse.js#L673-L688

This PR allows Papaparse to handle this situation and avoid downloading file content for many times.


papaparse.js
```diff
@@ -684,7 +682,10 @@ License: MIT
 				return;
 			}
 
-			this._finished = !this._config.chunkSize || this._start > getFileSize(xhr);
+			this._start += xhr.responseText.length;
+
```
pokoli: Could you avoid adding empty lines

t83714:
@pokoli Could you please have another look?
I have removed the empty lines and changes to the minified file.

pokoli:
@t83714 I've merged it. Thanks!

Issue #672, opened 2019-06-05
Both config skipEmptyLines: 'greedy' and step throw error


ubuntugx:
``` javascript
Papa.parse(event.target.files[0], {
    header: true,
    transformHeader: (val) => FIELDS_MAP[val],
    skipEmptyLines: 'greedy',  // Both config skipEmptyLines: 'greedy' and step throw error
    step: (results, parser) => {
      console.log('Row data:', results.data);
      console.log('Row errors:', results.errors);
      parser.resume();
    },
    complete: (results, file) => {
      console.log('Parsing complete:', results, file);
    }
});
```
`Uncaught TypeError: e.join is not a function`

``` javascript
return"greedy"===_.skipEmptyLines?""===e.join("") 
```
in this line

logol:
Can confirm this issue is still present with v5.0.0.
I have the problem that I cannot parse any file that has "newline at end of file" added e.g. when saving a CSV file under Linux using LibreOffice. I was hoping `skipEmptyLines: 'greedy'` solves this issue, but all attempts to parse a file now throw `e.join is not a function`

ConnorJBeck:
To add onto this, it appears setting `skipEmptyLines` to `true` does nothing when a step function is declared.

A simple test you can add to the `CUSTOM_TESTS` in `test-cases.js` to show this: 

```javascript
{
	description: "Data is correctly parsed with steps when skipping empty lines",
	expected: [['A', 'b', 'c'], ['d', 'E', 'f']],
	run: function(callback) {
		var data = [];
		Papa.parse('A,b,c\n\nd,E,f\n', {
			skipEmptyLines: true,
			step: function(results) {
				data.push(results.data);
			},
			complete: function() {
				callback(data);
			}
		});
	}
}
```

kaligrafy:
The problem is still there in 5.1.0:

skipEmptyLine greedy and step function triggers this error:

```node_modules/papaparse/papaparse.js:1126
return _config.skipEmptyLines === 'greedy' ? s.join('').trim() === '' : s.length === 1 && s[0].length === 0;
TypeError: s.join is not a function```

PR #714, opened 2019-10-06
Fix step callback function when skipping empty lines (#672)

seachicken:
Fixes https://github.com/mholt/PapaParse/issues/672#issuecomment-522758754


pokoli:
Thanks for your contribution!

Issue #713, opened 2019-10-03
Quote handling issue


jseter:
In my test 'questionably formatted' file, which is 2 MB the browser freezes, trying the parse the whole file as a single quote even though the quoting is properly formed where it is stuck.  This completely locks the browser tab at 100% cpu, never making it into the step function until it exhausts the entire file or runs out of memory queuing errors.

In debugging, the main issue i see is that:
```
// Check up to nextDelim or nextNewline, whichever is closest
var checkUpTo = nextNewline === -1 ? nextDelim : Math.min(nextDelim, nextNewline);
var spacesBetweenQuoteAndDelimiter = extraSpaces(checkUpTo);
``` 
nextDelim and nextNewline have not been updated at this point in the code, so if they are within the quoted text they will be less than quoteSearch, used in extraSpaces function.

File delimiter is "~", but the user saved the file using ',' and settings such that all the contents ended up in the first column.  the parser should have ended parsing the line at the first newline, but ends up in a loop trying to load the whole file as a single column.
ex.
```
"33~000114~000020333500000~003335~""CHAPPED LIPS FRESH SV    """,,,,
"33~000114~000200333500000~003335~""CHAPPED LIPS FRESH SV    """,,,,
"06~000196~000020777700000~007777~""WALL ZYGNA LIGHT SWAG CHS""",,,,
 ```

pokoli:
It seems that the issue is parsed due to using a bad delimiter. Is the file parsed correctly when using the right delimiter? I guess you should use the guessDelimiter functionality to avoid this kind of errors.

jseter:
Using a comma would behave better here of course, although would still be the wrong data since the file is malformed.  The main point though is that the after-quote handling in papa parse would not be correct in the case of either the delimiter being present in the quoted string or newline being in the quoted string.  Both of these cases are part of the reason quoting exists.

It should at a minimum update these indexes if the starting point, in this case `quoteSearch` is greater than or equal to their current value, which is a clear indication that those indexes are inside the quoted string and not after.
Additionally it is only checking for whitespace after the quoted text.  In this case there is unquoted text after the quote, which is lacking a defined behavior.  Since the RFC only specifies quoted/un-quoted for a field, i would assume that this case should have reported that as malformed and quit the quote parsing.  since it keeps searching for a closing quote followed by newline or delimiter it will   keep scanning until it hits the end of the file.

pokoli:
[...]

PR #719, opened 2019-10-16
#713 improve quote handling

jseter:
Fix substring being called with invalid indexes.  Added strict quote mode to treat malformed quote as unquoted field.  remove depreciated substr usage.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -699,7 +699,7 @@ License: MIT
 			if (contentRange === null) { // no content range, then finish!
 				return -1;
 			}
-			return parseInt(contentRange.substr(contentRange.lastIndexOf('/') + 1));
+			return parseInt(contentRange.substring(contentRange.lastIndexOf('/') + 1));
 		}
 	}
 	NetworkStreamer.prototype = Object.create(ChunkStreamer.prototype);
@@ -786,10 +786,17 @@ License: MIT
 		};
 		this._nextChunk = function()
 		{
+			var chunk;
 			if (this._finished) return;
 			var size = this._config.chunkSize;
-			var chunk = size ? remaining.substr(0, size) : remaining;
-			remaining = size ? remaining.substr(size) : '';
+			if(size) {
+				var limit = Math.min(remaining.length, Math.max(0, size));
+				chunk = remaining.substring(0, limit);
+				remaining = remaining.substring(limit);
+			} else {
+				chunk = remaining;
+				remaining = '';
+			}
 			this._finished = !remaining;
 			return this.parseChunk(chunk);
 		};
@@ -1092,7 +1099,7 @@ License: MIT
 		{
 			_paused = true;
 			_parser.abort();
-			_input = _input.substr(_parser.getCharIndex());
+			_input = _input.substring(_parser.getCharIndex());
 		};
 
 		this.resume = function()
@@ -1330,7 +1337,7 @@ License: MIT
 
 		function guessLineEndings(input, quoteChar)
 		{
-			input = input.substr(0, 1024 * 1024);	// max length 1 MB
+			input = input.substring(0, 1024 * 1024);	// max length 1 MB
 			// Replace all the text inside quotes
 			var re = new RegExp(escapeRegExp(quoteChar) + '([^]*?)' + escapeRegExp(quoteChar), 'gm');
 			input = input.replace(re, '');
@@ -1382,6 +1389,8 @@ License: MIT
 		var step = config.step;
 		var preview = config.preview;
 		var fastMode = config.fastMode;
+		var temp1;
+		var strictQuote = (temp1 = config.strictQuote) !== undefined ? temp1 : false;
 		var quoteChar;
 		/** Allows for no quoteChar by setting quoteChar to undefined in config */
 		if (config.quoteChar === undefined) {
@@ -1448,7 +1457,7 @@ License: MIT
 						cursor += newline.length;
 					else if (ignoreLastRow)
 						return returnable();
-					if (comments && row.substr(0, commentsLen) === comments)
+					if (comments && row.substring(0, commentsLen) === comments)
 						continue;
 					if (stepIsFunction)
 					{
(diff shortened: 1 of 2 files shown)


pokoli:
Hi, thanks for your proposal. 

Could you explain which is the objective of this PR and if which issue is fixing (if any).

jseter:
The primary objective is to address #713, where a malformed quoted input gets into a long loop that freezes the browser.  I added a strict quoting mode to fail as early as possible and treat the malformed input as an unquoted field.

Additionally while examining the code I found in cases where a quoted string contains the newline or delimiter, the current value of those indexes are stale when used by `extraSpaces()` which depends on the substring between the closing quote and the end of the field or record.  The default behavior of substring in this condition is to just flip its inputs which does not produce the desired result.
I added test cases to catch this problem.

Adjusted the index reported on quoting errors to represent the start of the quote, instead of the first character inside the quote.

limited the "deepEqual" undefined issue to address only the serialization of the result, instead of depending on JSON stringify obeying a specific ordering as the string compare would require.

Refactored to not use substr which is a legacy function and should be treated as depreciated.

jseter:
My test file still manages to freeze the browser, so more investigation is probably needed to handle this particular issue.

jseter:
browser is using minified file which isn't updated in my test.  using the non-minified file so far is working as expected.

pokoli:
> My test file still manages to freeze the browser, so more investigation is probably needed to handle this particular issue.

You should use `npm test` to corretcly run the tests using the proper version.

jseter:
yes, i used `npm test` during the development.  I did not submit until all tests passed.  What I meant is that I tried to use the library in an actual application against the file that raised the issue, but the minified file is not automatically generated during pack or test so the browser was using the old unchanged code.

pokoli:
> yes, i used `npm test` during the development. I did not submit until all tests passed. What I meant is that I tried to use the library in an actual application against the file that raised the issue, but the minified file is not automatically generated during pack or test so the browser was using the old unchanged code.

The minified file always includes the latest relase code and it's updated on the release process.

papaparse.js
```diff
@@ -1092,7 +1099,7 @@ License: MIT
 		{
 			_paused = true;
 			_parser.abort();
-			_input = _input.substr(_parser.getCharIndex());
+			_input = _input.substring(_parser.getCharIndex());
```
pokoli: Why changing this?

papaparse.js
```diff
@@ -1330,7 +1337,7 @@ License: MIT
 
 		function guessLineEndings(input, quoteChar)
 		{
-			input = input.substr(0, 1024 * 1024);	// max length 1 MB
+			input = input.substring(0, 1024 * 1024);	// max length 1 MB
```
pokoli: Idem, why changing?

papaparse.js
```diff
@@ -1382,6 +1389,8 @@ License: MIT
 		var step = config.step;
 		var preview = config.preview;
 		var fastMode = config.fastMode;
+		var temp1;
+		var strictQuote = (temp1 = config.strictQuote) !== undefined ? temp1 : false;
```
pokoli: This line is a litte bit strange and hard to understand. Could you explain why it is needed?

papaparse.js
```diff
@@ -1731,7 +1774,7 @@ License: MIT
 			}
 
 			/** Gets the delimiter character, which is not inside the quoted field */
-			function getNextUnqotedDelimiter(nextDelim, quoteSearch, newLine) {
+			function getNextUnquotedDelimiter(nextDelim, quoteSearch, newLine) {
```
pokoli: Mmm, I will be happy to merge this as a separate PR as it does not seem releated.

tests/test-cases.js
```diff
@@ -22,6 +22,61 @@ try {
 	XHR_ENABLED = true;
 } catch (e) {} // safari, ie
 
+var __substring = String.prototype.substring;
```
pokoli: Why is this all stuff neeeded?

tests/test-cases.js
```diff
@@ -1475,7 +1580,7 @@ describe('Parse Tests', function() {
 			if (test.expected.meta) {
 				assert.deepEqual(actual.meta, test.expected.meta);
 			}
-			assert.deepEqual(JSON.stringify(actual.errors), JSON.stringify(test.expected.errors));
+			assert.deepEqual(JSON.parse(JSON.stringify(actual.errors)), test.expected.errors);
```
pokoli: Why changing this?

tests/test-cases.js
```diff
@@ -246,14 +317,14 @@ var CORE_PARSER_TESTS = [
 				"code": "InvalidQuotes",
 				"message": "Trailing quote on quoted field is malformed",
 				"row": 0,
-				"index": 3
+				"index": 2
```
pokoli: Why changing the index?

papaparse.js
jseter: Refactored to not use substr which is a legacy function and should be treated as depreciated.

papaparse.js
jseter: instead of fetching strictQuote from config once to test it for being undefined and again to use the value (which could have changed between the two points if the object installed a getter method, it saves it to a temporary local.
```
var strictQuote;

var temp1 = config.strictQuote;
if(temp1 !== undefined) {
    strictQuote = temp1;
} else {
    strictQuote = false; // default value
}
```

tests/test-cases.js
jseter: it allows the unit test to fail if substring is called with bad inputs, which is the first issue being addressed.

tests/test-cases.js
jseter: as mentioned in the explanation, the index should represent the beginning of the quote or column and not inside the quoted string.

tests/test-cases.js
jseter: as covered in the explanation, JSON does not guarantee ordering or the properties within an object so doing a string compare of two JSON strings is not a good practice.  The goal of the JSON hack is to eliminate the `undefined` properties which deepEqual strictly would fail this on.  I have addressed in a separate branch making PapaParse not produce the undefined value so that this wouldn't be needed in that branch.
This modification keeps the value assertion `test.expected.errors` left untampered while using the JSON workaround for its intended purpose and avoiding the ordering problem.

papaparse.js
pokoli: Ok, could you please move this to a separate PR which does not modify the test suite?

jseter:
Will break into separate PR's to simplify.

PR #729, opened 2019-10-24
Update papaparse.min.js

MrSam:
latest code


pokoli:
Thanks for your pull request but this is not necessary as the minified file is updated on each realease.

MrSam:
Hmm, but papaparse.js is 3hours old and the min is 1 month old ?

pokoli:
Yes because latests relesase is one month old

PR #740, opened 2019-11-15
remove limitation on `newline` option

### LINKED ISSUE #737: random newline character
Why is paparse limiting the newline character to only \r, \n and \r\n ?
To simplify communication between OSes, I have gotten into the habit of delimiting new lines with other characters than the actual new lines characters.

elChapoSing:
fixes #737

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -1388,12 +1388,13 @@ License: MIT
 		// Unpack the config object
 		config = config || {};
 		var delim = config.delimiter;
-		var newline = config.newline;
 		var comments = config.comments;
 		var step = config.step;
 		var preview = config.preview;
 		var fastMode = config.fastMode;
 		var quoteChar;
+		var newline;
+		
 		/** Allows for no quoteChar by setting quoteChar to undefined in config */
 		if (config.quoteChar === undefined) {
 			quoteChar = '"';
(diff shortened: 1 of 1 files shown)


elChapoSing:
not sure why the check fails...
@pokoli

papaparse.js
```diff
@@ -1419,10 +1419,6 @@ License: MIT
 			|| Papa.BAD_DELIMITERS.indexOf(comments) > -1)
 			comments = false;
 
-		// Newline must be valid: \r, \n, or \r\n
-		if (newline !== '\n' && newline !== '\r' && newline !== '\r\n')
```
pokoli: Instead of removing the test you should test that newLine is undefined. In this case the default value `\n` should be used.

This is the reason why all tests are failing. 

P.S: You can run the tests on your computer with `npm test` command.

elChapoSing:
the only errors I see when running `npm test` are linebreaks
```
 Expected linebreaks to be 'LF' but found 'CRLF'  linebreak-style
```

And I have no idea how to fix them.

elChapoSing:
Any love for that PR ? @pokoli

pokoli:
You should configure your editor to use LF as line breaks insteaad of CRLF.
On linux there is a tool called dos2unix which can be used to fix this problem.

elChapoSing:
sorry but I'm not modifying my configuration for this.
Simple enough to replicate my changes on your side if you think it makes sense.

PR #742, opened 2019-11-15
Add checking for null when creating `global` object

antoinegoutagny:
For some reasons, `self` is null in some contexts. Adding a null check prevents issue with this.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -37,9 +37,9 @@ License: MIT
 		// but without using `eval` (which is disabled when
 		// using Content Security Policy).
 
-		if (typeof self !== 'undefined') { return self; }
-		if (typeof window !== 'undefined') { return window; }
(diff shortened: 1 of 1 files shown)


pokoli:
Could you elaborate in which contexts self is null?

antoinegoutagny:
I'm afraid I don't know. We use PapaParse in the browser in a fairly large legacy Polymer 1.0 application, and we sometimes get an error because `global` is null. We tracked it down to `self` being null. It's probably another part of the app that messes with `self`, but I can't say where.
Anyway, it doesn't hurt to have a more robust check when initializing PapaParse.

pokoli:
I do no think we should add extra checking without a proper reason to add it. 
Normally self, window and global are correctly defined on their respective scopes, so probably this is not an issue on PapaParse but on the code which executes it.

Issue #736, opened 2019-11-14
Chunking broken with pause() and resume()


duclet:
Hi, I am trying to use streaming with chunk that inside of it make use of pause and resume. I'm noticing that after the pause and resume, the data of the chunk is incorrect. Essentially, the line where it concatenate the with the reader result, the partial line already contains a portion of the reader result. This can be easily reproduced with the "Local Files(s)" demo on the web site using the Large file. You just need to change the config to be the following:

```javascript
                config: {
					chunk: function(results, parser) {
						parser.pause();
						parser.resume();
					},
				},
```
After pausing and when `resume` is called, when the code makes it to the `parseChunk` method, you'll notice that the contents of `this._partialLine` contains the argument `chunk`.

I've dig a little bit into the code but since it is very large, I'm not sure fully what is going on. However, on pause, the line `_input = _input.substring(_parser.getCharIndex());` doesn't seem like it is necessary. If I set `_input` to just an empty string, issue is resolved.

pokoli:
Could you please add a PR with your fix and a test case that reproduces your problem? 
It will be great to have this as part of papaparse

duclet:
@pokoli - I can definitely create a PR but not with the fix I mentioned above. The reason being is that it was just me temporarily testing it out to see if it fixes the issue I was running into. Considering the line added over 5 years ago, I find it a little hard to believe no one has ran into this issue. That is why the line was there in the first place. Unfortunately, I don't have that knowledge so I can't commit any changes yet. I would like to take more time to figure out the entire flow and problem. That being said, if the original author can chime in for why it exists, that would be very helpful.

pokoli:
@duclet Don't worry, just create the PR with the fix, a test that reproduces the issue. 

With all of this information we can decide if your fix is correct or if there is any side-effect.

duclet:
@pokoli - Created the PR: https://github.com/mholt/PapaParse/pull/745

PR #745, opened 2019-11-16
Fixes bug #736

duclet:


papaparse.js
```diff
@@ -682,7 +682,10 @@ License: MIT
 				return;
 			}
 
-			this._start += xhr.responseText.length;
+			// The length of the responseText doesn't corresponds to the actual size we requested due to potential encoding
+			// of some characters that has more than 1 byte. As such, better to just increment the start index by the chunk
+			// size if it was given
+			this._start += this._config.chunkSize ? this._config.chunkSize : xhr.responseText.length;
```
duclet: Personally I think the better solution here is to use the TextEncoder to properly get the correct byte size. But I'm not sure if it is available in all the browsers you support plus that would also mean we would need to parse the content to get the right size. Now sure if all that is worth it so going with this.

papaparse.js
```diff
@@ -1100,7 +1103,10 @@ License: MIT
 		{
 			_paused = true;
 			_parser.abort();
-			_input = _input.substring(_parser.getCharIndex());
+
+			// If it is streaming via "chunking", the reader will start appending correctly already so no need to substring,
+			// otherwise we can get duplicate content within a row
+			_input = isFunction(_config.chunk) ? "" : _input.substring(_parser.getCharIndex());
```
duclet: My original recommendation in the bug ticket was to just set this to an empty string or remove the usage of `_input` all together. Turns out `step` needs this while `chunk` doesn't.

tests/test-cases.js
```diff
@@ -1897,6 +1897,114 @@ var CUSTOM_TESTS = [
 			});
 		}
 	},
+	{
+		description: "Pause and resume works for chunks with NetworkStreamer (for Bug #736)",
```
pokoli: I do not think is needed to include the bug # in the test. It can be directly obtained from git history

tests/test-cases.js
```diff
@@ -1897,6 +1897,114 @@ var CUSTOM_TESTS = [
...
+		disabled: !XHR_ENABLED,
+		timeout: 30000,
+		expected: ["SembCorp Industries Ltd", "Singapore", "SCIL.SI", "10%", "Yes", "0.30%"],
+		run: function(callback) {
+			var chunkNum = 0;
+			Papa.parse(BASE_PATH + "duplicate.csv", {
```
pokoli: Do we really need a new test file? 
Could you please use long-sample.csv? If you need some special character just edit the file and add them.

tests/test-cases.js
```diff
@@ -1897,6 +1897,114 @@ var CUSTOM_TESTS = [
...
+				download: true,
+				chunkSize: 250000,
```
pokoli: I think we can use a lower chunkSize just for testing.

tests/test-cases.js
```diff
@@ -1897,6 +1897,114 @@ var CUSTOM_TESTS = [
...
+
+					if (chunkNum === 2) {
+						callback(results.data[0]);
+						return;
+					}
+
+					parser.resume();
+				},
+				complete: function() {
+					callback(new Error("Should have more than 2 chunks"));
```
pokoli: I think you are missing a negation on the error message: "Should not have more than 2 chuncks"

tests/test-cases.js
```diff
@@ -1897,6 +1897,114 @@ var CUSTOM_TESTS = [
...
+				callback(err);
+				return;
+			}
+		}
+	},
+	{
+		description: "Pause and resume works for chunks with StringStreamer (for Bug #736)",
+		disabled: !XHR_ENABLED,
+		timeout: 30000,
+		// For those wondering why this is different than the two above, reading by byte size isn't exactly the same as a
```
pokoli: I think we can simplify the comment to: "Test also with string as byte size may be diferent"

tests/test-cases.js
```diff
@@ -1897,6 +1897,114 @@ var CUSTOM_TESTS = [
...
+	},
+	{
+		description: "Pause and resume works for chunks with FileStreamer (for Bug #736)",
+		disabled: !XHR_ENABLED,
+		timeout: 30000,
+		expected: ["SembCorp Industries Ltd", "Singapore", "SCIL.SI", "10%", "Yes", "0.30%"],
+		run: function(callback) {
+			var chunkNum = 0;
+
+			// A little bit of a hack but this allows us to test the FileStreamer for local files. Essentially, this uses the
```
pokoli: I do not think this comment is required

papaparse.js
```diff
@@ -682,7 +682,10 @@ License: MIT
 				return;
 			}
 
-			this._start += xhr.responseText.length;
+			// The length of the responseText doesn't corresponds to the actual size we requested due to potential encoding
```
pokoli: It can be simplified to: "Use chunckSize as it may be a diference on reponse lentgh due to characters with more than 1 byte.

tests/test-cases.js
duclet: I'll update messaging since it isn't right either way.

papaparse.js
duclet: @pokoli - I guess I can also use `this._start += +xhr.getResponseHeader('Content-Length');` from another bug request though I'm a little hesitant to use headers since it is dependent on the server which may or may not provide the information. Thoughts?

tests/verylong-sample.csv
```diff
@@ -1,7 +1,7 @@
 placeholder,meaning of life,TLD
 Lorem ipsum dolor sit,42,ABC
 Etiam a dolor vitae est vestibulum,84,DEF
-Lorem ipsum dolor sit,42,ABC
+"Lorem ipsum dolor sit",42,ABC
```
duclet: Turns out that quote around the string cause the problem.

papaparse.js
pokoli: I agree that we should avoid using headers if they are optionals. So current implementation seems ok.

pokoli:
It seems there is something wrong with the test as they are not failing when commenting the patch on papaparse.js

I'm not sure this is releated to this PR but I will like to get this fixed it before merging

duclet:
> It seems there is something wrong with the test as they are not failing when commenting the patch on papaparse.js
> 
> I'm not sure this is releated to this PR but I will like to get this fixed it before merging

@pokoli - You should change line 1107 to `_input = _input.substring(_parser.getCharIndex());` to restore the previous behavior, not commenting it out. That should fail the tests.

pokoli:
Yes, that's what I've done and tests are not failing

duclet:
@pokoli - Odd, it is definitely failing for me. Curious, which version of Node are you using?

pokoli:
$ node --version
v12.13.0

duclet:
@pokoli - I tried it with that version and it definitely failed:

```
...
  Custom Tests
    ✓ Pause and resume works (Regression Test for Bug #636) (8518ms)
    1) Pause and resume works for chunks with NetworkStreamer
[Error: AssertionError]
    2) Pause and resume works for chunks with FileStreamer
[Error: AssertionError]
    3) Pause and resume works for chunks with StringStreamer
[Error: AssertionError]
    ✓ Complete is called with all results if neither step nor chunk is defined
    ✓ Step is called for each row
    ✓ Data is correctly parsed with steps
    ✓ Data is correctly parsed with steps (headers)
    ✓ Data is correctly parsed with steps and worker (headers)
    ✓ Data is correctly parsed with steps and worker
    ✓ Data is correctly parsed with steps when skipping empty lines
    ✓ Step is called with the contents of the row
    ✓ Step is called with the last cursor position
    ✓ Step exposes cursor for downloads
    ✓ Step exposes cursor for chunked downloads
    ✓ Step exposes cursor for workers
    ✓ Chunk is called for each chunk
    ✓ Chunk is called with cursor position
    ✓ Chunk functions can pause parsing
    ✓ Chunk functions can resume parsing (500ms)
    ✓ Chunk functions can abort parsing
    ✓ Step exposes indexes for files
    ✓ Step exposes indexes for chunked files
    ✓ Quoted line breaks near chunk boundaries are handled
    ✓ Step functions can abort parsing
    ✓ Complete is called after aborting
    ✓ Step functions can pause parsing
    ✓ Step functions can resume parsing (501ms)
    ✓ Step functions can abort workers
    ✓ beforeFirstChunk manipulates only first chunk
    ✓ First chunk not modified if beforeFirstChunk returns nothing
    ✓ Should correctly guess custom delimiter when passed delimiters to guess.
    ✓ Should still correctly guess default delimiters when delimiters to guess are not given.


  219 passing (10s)
  3 failing

  1) Custom Tests
       Pause and resume works for chunks with NetworkStreamer:
     Uncaught Error: Uncaught AssertionError: expected [ Array(3) ] to deeply equal [ Array(3) ] (http://localhost:8071/node_modules/chai/chai.js:239)
  

  2) Custom Tests
       Pause and resume works for chunks with FileStreamer:
     Uncaught Error: Uncaught AssertionError: expected [ Array(3) ] to deeply equal [ Array(3) ] (http://localhost:8071/node_modules/chai/chai.js:239)
  

  3) Custom Tests
       Pause and resume works for chunks with StringStreamer:
     Uncaught Error: Uncaught AssertionError: expected [ Array(3) ] to deeply equal [ Array(3) ] (http://localhost:8071/node_modules/chai/chai.js:239)
```

For the record, this is the diff:
```
--- a/papaparse.js
+++ b/papaparse.js
@@ -1104,7 +1104,8 @@ License: MIT
 
             // If it is streaming via "chunking", the reader will start appending correctly already so no need to substring,
             // otherwise we can get duplicate content within a row
-            _input = isFunction(_config.chunk) ? "" : _input.substring(_parser.getCharIndex());
+            // _input = isFunction(_config.chunk) ? "" : _input.substring(_parser.getCharIndex());
+            _input = _input.substring(_parser.getCharIndex());
         };
```

PR #769, opened 2020-03-03
Using self instead of this to preserve binding.

varunsharma27:
This should be `self.resume` instead of `this.resume` as `this` refers to the `window` object when the callback occurs. Meaning, it would fail to do the next recursive call if the need arose.

Do you have any testing strategy in mind for such a thing? Do we need a test here?


pokoli:
Merged it! Thanks!

PR #789, opened 2020-04-30
Add timeout option to download

### LINKED ISSUE #786: timeout for download
The download option currently provides no way to set a timeout, so when a server misbehaves, it can take a long time before an error callback arrives.

The download timoeut should be configurable.

demetris-manikas:
Fixes #786
Could not locate where to update the docs.
Thanks a lot. PapaParse rocks!

### DIFF
--- a/docs/docs.html
+++ b/docs/docs.html
@@ -136,6 +136,7 @@
 					<div class="grid-50">
 						<pre><code class="language-javascript">Papa.parse(url, {
 	download: true,
+	downloadTimeout: 5000,
 	// rest of config ...
 })</code></pre>
 					</div>
--- a/papaparse.js
+++ b/papaparse.js
@@ -328,7 +328,7 @@ License: MIT
 				return;
 
 			if (typeof _config.delimiter === 'string'
-                && !Papa.BAD_DELIMITERS.filter(function(value) { return _config.delimiter.indexOf(value) !== -1; }).length)
+				&& !Papa.BAD_DELIMITERS.filter(function(value) { return _config.delimiter.indexOf(value) !== -1; }).length)
 			{
 				_delimiter = _config.delimiter;
 			}
@@ -589,7 +589,6 @@ License: MIT
 		}
 	}
 
-
 	function NetworkStreamer(config)
 	{
 		config = config || {};
@@ -643,6 +642,9 @@ License: MIT
 			}
 
 			xhr.open(this._config.downloadRequestBody ? 'POST' : 'GET', this._input, !IS_WORKER);
+			xhr.timeout = this._config.downloadTimeout;
+			xhr.ontimeout = bindFunction(function() { this._chunkError('timeout'); }, this);
+
 			// Headers can only be set when once the request state is OPENED
 			if (this._config.downloadRequestHeaders)
 			{
@@ -1397,6 +1399,7 @@ License: MIT
 		var step = config.step;
 		var preview = config.preview;
 		var fastMode = config.fastMode;
+		var downloadTimeout;
 		var quoteChar;
 		/** Allows for no quoteChar by setting quoteChar to undefined in config */
 		if (config.quoteChar === undefined) {
@@ -1427,6 +1430,16 @@ License: MIT
 		if (newline !== '\n' && newline !== '\r' && newline !== '\r\n')
 			newline = '\n';
 
+		if (config.downloadTimeout) {
+			downloadTimeout = parseInt(config.downloadTimeout);
+			if (!isNaN(downloadTimeout)) {
+				config.downloadTimeout = downloadTimeout;
+			} else {
+				throw new Error('Config downloadTimeout value (' + config.downloadTimeout + ') not parsable by parseInt(val).');
+			}
+		} else {
+			config.downloadTimeout = 0;
+		}
 		// We're gonna need these at the Parser scope
 		var cursor = 0;
 		var aborted = false;
@@ -1680,9 +1693,9 @@ License: MIT
 			}
 
 			/**
-             * checks if there are extra spaces after closing quote and given index without any text
-             * if Yes, returns the number of spaces
-             */
+			 * checks if there are extra spaces after closing quote and given index without any text
+			 * if Yes, returns the number of spaces
+			 */
 			function extraSpaces(index) {
 				var spaceLength = 0;
 				if (index !== -1) {
--- a/tests/test-cases.js
+++ b/tests/test-cases.js
@@ -1518,8 +1518,6 @@ describe('Parse Tests', function() {
 	}
 });
 
-
-
 // Tests for Papa.parse() that involve asynchronous operation
 var PARSE_ASYNC_TESTS = [
 	{
@@ -1580,7 +1578,21 @@ var PARSE_ASYNC_TESTS = [
 			data: [['A','B','C'],['X','Y','Z']],
 			errors: []
 		}
-	}
+	},
+	{
+		description: "Simple download with int timeout timed out",
+		input: BASE_PATH + "verylong-sample.csv",
+		config: {
+			download: true,
+			downloadTimeout: 1
+		},
+		disabled: !XHR_ENABLED,
+		expected: {
+			exception: {
+				"message": "timeout"
+			}
+		}
+	},
 ];
 
 describe('Parse Async Tests', function() {
@@ -1595,7 +1607,12 @@ describe('Parse Async Tests', function() {
 			};
 
 			config.error = function(err) {
-				throw err;
+				if (!test.expected.exception) {
+					throw err;
+				} else {
+					assert.equal(err.message, test.expected.exception.message);
+					done();
+				}
 			};
 
 			Papa.parse(test.input, config);


pokoli:
Hi @demetris-manikas, 

First of all thanks for your Pull Request. 

The docs are included in the docs/ folder of the source code repository. There is a section where all the options are explained. You should add the timeout option with a brief description of what it does.

I think it makes sense to describe it after the downaload option (as they are related).

papaparse.js
```diff
@@ -643,6 +644,20 @@ License: MIT
 			}
 
 			xhr.open(this._config.downloadRequestBody ? 'POST' : 'GET', this._input, !IS_WORKER);
+			if (this._config.download && this._config.download.timeout) {
```
pokoli, line 647: I do not think using a nested option is a good idea.

Better to name it downloadTimeout

papaparse.js
```diff
@@ -643,6 +644,20 @@ License: MIT
...
+				timeout = this._config.download.timeout;
+				timeout = typeof timeout === 'number' ?
```
pokoli, line 649: I think it's simpler to directly call parseInt on the option and only use if is set.

demetris-manikas:
As you requested :)

papaparse.js
demetris-manikas: Done

papaparse.js
demetris-manikas: Done

papaparse.js
```diff
@@ -643,6 +644,15 @@ License: MIT
 			}
 
 			xhr.open(this._config.downloadRequestBody ? 'POST' : 'GET', this._input, !IS_WORKER);
+			if (this._config.downloadTimeout) {
+				timeout = parseInt(this._config.downloadTimeout);
+				if (timeout) {
+					xhr.timeout = timeout;
+					xhr.ontimeout = bindFunction(function() { this._chunkError('timeout'); }, this);
+				} else {
+					this._chunkError('invalid timeout');
```
AndreasHassing, line 653: Consider helping the user here. "Why was my timeout invalid?"

AndreasHassing: It may be enough to just tell them what they tried to use and maybe also what type that was, and what type was expected.

tests/test-cases.js
```diff
@@ -1501,6 +1501,39 @@ describe('Parse Tests', function() {
...
+	{
+		description: "Simple download with int timeout timed out",
+		input: BASE_PATH + "verylong-sample.csv",
+		config: {
+			download: true,
+			downloadTimeout: 1
+		},
+		disabled: !XHR_ENABLED,
+		expected: {
+			error: "timeout"
```
AndreasHassing, line 1514: This doesn't look right compared to all the other test cases.

Shouldn't it be something like(?):
```javascript
expected: {
    data: [],
    errors: ["timeout"] // not even sure about this, there doesn't seem to be any other tests for exceptions
}
```

tests/test-cases.js
```diff
@@ -1501,6 +1501,39 @@ describe('Parse Tests', function() {
 	}
 });
 
+var PARSE_ASYNC_ERROR_TESTS = [
```
AndreasHassing, line 1504: Why couldn't this test case go into the `PARSE_ASYNC_TESTS` test cases?

demetris-manikas:
@AndreasHassing Thanks for your input. I 'll try to get this ready in the next few days.

docs/docs.html
```diff
@@ -136,13 +136,15 @@ <h5 id="remote-files">Parse remote file</h5>
...
 	download: true,
+	downloadTimeout: 5000,
 	// rest of config ...
 })</code></pre>
 					</div>
 					<div class="grid-50">
 						<ul>
 							<li><code>url</code> is the path or URL to the file to download.</li>
 							<li>The second argument is a <a href="#config">config object</a> where <code>download: true</code> is set.</li>
+							<li>The third argument is a <a href="#config">config object</a> where <code>downloadTimeout: 5000 (millis)</code> is set.</li>
```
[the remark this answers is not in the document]
demetris-manikas: Fixed.

docs/docs.html
```diff
@@ -136,13 +136,15 @@ <h5 id="remote-files">Parse remote file</h5>
 					<div class="grid-50">
 						<pre><code class="language-javascript">Papa.parse(url, {
 	download: true,
+	downloadTimeout: 5000,
```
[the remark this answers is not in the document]
demetris-manikas: XMLHttpRequest calls it "timeout". I prefer to use the same name.

papaparse.js
```diff
@@ -598,6 +598,7 @@ License: MIT
 		ChunkStreamer.call(this, config);
 
 		var xhr;
+		var timeout;
```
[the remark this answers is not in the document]
demetris-manikas: XMLHttpRequest calls it "timeout". I prefer to use the same name.

papaparse.js
```diff
@@ -643,6 +644,15 @@ License: MIT
 			}
 
 			xhr.open(this._config.downloadRequestBody ? 'POST' : 'GET', this._input, !IS_WORKER);
+			if (this._config.downloadTimeout) {
+				timeout = parseInt(this._config.downloadTimeout);
+				if (timeout) {
```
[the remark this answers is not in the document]
demetris-manikas: Fixed

papaparse.js
demetris-manikas: Updated message to be more descriptive.

tests/test-cases.js
demetris-manikas: No.

tests/test-cases.js
demetris-manikas: No.

papaparse.js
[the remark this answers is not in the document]
AndreasHassing: Personally, I'm not a fan of exposing bad naming from inner components to the outside world. Just because `XMLHttpRequest` has ambiguous names, does not mean that we have to 😊.

Just my 2 cent.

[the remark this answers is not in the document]
AndreasHassing: Aside: seeing as this variable is not used anywhere but within the if-block of `if (this._config.downloadTimeout) {`, consider moving the variable into there.

AndreasHassing:
Only had a minor comment - looks good @demetris-manikas, looking forward to having this in 🥳.

docs/docs.html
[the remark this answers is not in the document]
pokoli: I think the naming is correct. Addding the unit will only makes it longer with no extra information

papaparse.js
```diff
@@ -643,6 +644,15 @@ License: MIT
 			}
 
 			xhr.open(this._config.downloadRequestBody ? 'POST' : 'GET', this._input, !IS_WORKER);
+			if (this._config.downloadTimeout) {
+				timeout = parseInt(this._config.downloadTimeout);
+				if (!isNaN(timeout)) {
+					xhr.timeout = timeout;
+					xhr.ontimeout = bindFunction(function() { this._chunkError('timeout'); }, this);
+				} else {
+					this._chunkError('Config downloadTimeout value (' + this._config.downloadTimeout + ') not parsable by parseInt(val).');
```
pokoli, line 653: I'm not sure if it really makes sense to raise an error or just ignore the non Numeric timeout.

tests/test-cases.js
pokoli: Yes, please add this to the PARSE_ASYNC_TESTS

papaparse.js
AndreasHassing: I disagree. Why shouldn't the user be informed of misuse of the API?

pokoli: Maybe but using a wrong timeout will not cause any parsing error.

docs/docs.html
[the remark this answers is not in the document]
AndreasHassing: I believe it's incorrect to say that it doesn't add any information: it tells you exactly how to use the API at the call site instead of forcing the consumer to look up documentation or, worse, the source code.

As an example, what do you think the following API does?:
```javascript
Cache.prototype.setEvictionTime = function(timeout) { ... } ;
// what unit is timeout?

// vs

Cache.prototype.setEvictionTime = function (timeoutMs) { ... };
// Alright! I know what the parameter is 🥳
```

papaparse.js
AndreasHassing: You're right - but the user would unknowingly get incorrect behavior from the library (because `timeout` is not set). Being silently in an incorrect state is worse than failing completely, in my opinion.

pokoli: Then probably better to raise and exception but not a parsing one.

AndreasHassing: Ah, right - absolutely! I agree with you on that.

docs/docs.html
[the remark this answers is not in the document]
demetris-manikas: There is more to it. The config value is passed to parseInt (I tried to use radix 10 but the tests fail). The [parseInt](https://www.w3schools.com/jsref/jsref_parseint.asp) method also parses octal and hexadecimal values. So as things are now calling it timeoutMs is not correct in my opinion.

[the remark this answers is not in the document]
AndreasHassing: @demetris-manikas how does the number-base affect what unit it is?
2ms = 2 milliseconds
0x10ms = 2 milliseconds, also.

demetris-manikas:
True. Just realized and deleted my comment.

papaparse.js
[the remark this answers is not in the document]
demetris-manikas: Fixed

tests/test-cases.js
demetris-manikas: Fixed

papaparse.js
demetris-manikas: @pokoli I moved the check into Parser (where it  belongs) and I saw that in other cases no error is raised. So I believe that to keep things uniform no Error should be raised in case of invalid input. My personal opinion is that a user must always be informed of invalid values in configuration. Anyway it's your call.

pokoli: @demetris-manikas If you prefer to inform of invalid configuration it should not be a parse execption and we should warn for all invalid operations. But this is out of the scope of this PR (i prefer to keep them minumum)

Feel free to create a separate PR that adds it for all the current arguments.

demetris-manikas:
@pokoli Please review the latest code. An Error is thrown if config value is not parsable.

PR #790, opened 2020-05-01
add test to check for empty field in the begining

demetris-manikas:
Tests for #685.
I was not able to reproduce the issue but a test is always better to have.
I also checked the proposed failing cases on [https://www.papaparse.com/demo](https://www.papaparse.com/demo) and they are parsed as expected.
I am using PapaParse for a long time now and I would like to help taking this 109 issues down :).


pokoli:
Thanks for contributing back to the library ;)

Issue #804, opened 2020-06-19
Guessing wrong delimiter


Klingseis:
In #687 there was new guessing of right delimiter added. Right now I'm facing the issue, that it predicts "," (Comma) instead of ";" (semicolon).

The reason for this seems to be the final checking if the previewed delimiter is better or not:
`if ((typeof bestDelta === 'undefined' || delta <= bestDelta)
					&& (typeof maxFieldCount === 'undefined' || avgFieldCount > maxFieldCount) && avgFieldCount > 1.99)`

~The delta for semicolon (0) is smaller than bestDelta (5) but the avgFieldCount equals maxFieldCount. For me this looks like you could assume that this delimiter is better, couldn't you?
Maybe I am wrong, but wouldn't it be better to compare avgFieldCount >= maxFieldCount if the delta is smaller?~

[testGeometries.txt](https://github.com/mholt/PapaParse/files/4805708/testGeometries.txt): File to reproduce similar issue.

One correction from my side: My thought above of comparing avgFieldCount >= maxFieldCount if delta is smaller doesn't work.
In attached file delimiter is semicolon, papaparse guesses comma:

comma delta: 4, comma avgFieldCount: 4.6
semicolon delta: 0, semicolon avgFieldCount: 3

Klingseis:
[testGeometries.txt](https://github.com/mholt/PapaParse/files/4805607/testGeometries.txt): File to reproduce similar issue.

One correction from my side: My first thought of comparing avgFieldCount >= maxFieldCount if delta is smaller doesn't work. 
In attached file delimiter is semicolon, papaparse guesses comma:
- comma delta: 4, comma avgFieldCount: 4.6
- semicolon delta: 0, semicolon avgFieldCount: 3

Githamza:
I have the same issue, any solution ?

Klingseis:
I use PapaParse together with GUI. If a parsing error occurs or the result is only one column, a fieldset will be displayed to the user where he can type in a delimiter by his own and parse the file again.

Klingseis:
Will close this one. Only solution right now is to set the delimiter directly instead of guessing it by PapaParse.

PR #805, opened 2020-06-21
Add further check to guessDelimiter if its better (#804)

Klingseis:
In guessDelimiter()-function when checking if delimiter might be better, I added the check `delta < bestDelta` in case of avgFieldCount isn't bigger than maxFieldCount. Therefore I also added a test case to verify.

In my opinion you can say that if delta is smaller than bestDelta the delimiter is more appropriate even if avgFieldCount is smaller than maxFieldCount in case avgFieldCount is bigger than 1.99. I had a file where it guessed comma instead of semicolon, see #804. 
The reason for this was that in this file there are more commas than semicolons, so the avgFieldCount was higher for comma than for semicolon although the delta was zero for semicolon but four for comma.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -1335,7 +1335,7 @@ License: MIT
 					avgFieldCount /= (preview.data.length - emptyLinesCount);
 
 				if ((typeof bestDelta === 'undefined' || delta <= bestDelta)
-					&& (typeof maxFieldCount === 'undefined' || avgFieldCount > maxFieldCount) && avgFieldCount > 1.99) {
+					&& ((typeof maxFieldCount === 'undefined' || avgFieldCount > maxFieldCount) || delta < bestDelta) && avgFieldCount > 1.99) {
 					bestDelta = delta;
 					bestDelim = delim;
 					maxFieldCount = avgFieldCount;
--- a/tests/test-cases.js
+++ b/tests/test-cases.js
@@ -4,7 +4,7 @@ if (typeof module !== 'undefined' && module.exports) {
 	chai = require('chai');
 	Papa = require('../papaparse.js');
 }
-var fs = require('fs');
+
 var assert = chai.assert;
 
 var BASE_PATH = (typeof document === 'undefined') ? './' : document.getElementById('test-cases').src.replace(/test-cases\.js$/, '');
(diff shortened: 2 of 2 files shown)


papaparse.js
```diff
@@ -1335,7 +1335,7 @@ License: MIT
 					avgFieldCount /= (preview.data.length - emptyLinesCount);
 
 				if ((typeof bestDelta === 'undefined' || delta <= bestDelta)
-					&& (typeof maxFieldCount === 'undefined' || avgFieldCount > maxFieldCount) && avgFieldCount > 1.99) {
+					&& ((typeof maxFieldCount === 'undefined' || avgFieldCount > maxFieldCount) || delta < bestDelta) && avgFieldCount > 1.99) {
```
pokoli, line 1338: This is already tested in the line avobe. So for me this condition is redundant.

Klingseis: The problem is, that if delta is smaller or equal bestDelta, avgFieldCount needs to be bigger than maxFieldCount which might be not the case if e.g. comma appears more often but delimiter is another sign. Therefore I added the case that if avgFieldCount isn't bigger but delta is smaller, the delimiter is more accurate.

pokoli: This sound so much complicated and I prefer to keep it simple (or at least what it is now)

I do not like so much the current solution as we will always end up with a case where is not supported which requires PapaParse to be modified. 

So I'm wondering if it won't be better to allow the end user to pass a function that accepts the current parameters and returns true or false to determine if the delimiter is accepted or not.

PapaParse should have a default function (which should be callable from user custom functions) that will be used if no user defined function.

What do you think?

Klingseis: I'm not sure if we talk about the same topic. 
What do you mean with "current solution": mine or the one in repo right now?

I'm not talking about accepted delimiters or not, I just had a file where it guessed the wrong delimiter because in it (see #804 for example file) there were more commas included (due to geometry objects which are separated by comma) than semicolons (which in this case is delimiter as you can see in header).
The reason for this was that avgFieldCount for comma was of course higher because it ended in more fileds. But the delta was 4 because the header was considered as one field.

So when guessing the right delimiter I think we could approach that a smaller delta is always better if avgFieldCount is bigger than 2. 
The comparison avgFieldCount > maxFieldCount for me makes only sense if deltas are equal.
So maybe the comparison suggested in my comment above is less redundant and does the same job.

But maybe I made an error in reasoning, I'm not sure. I even didn't test the approach in comment above yet.

pokoli: "current solution" means the current one and it is exactly the same design as your proposal.
What I talk about is that we had so much request to adapt the delimiter guessing functionality to specific case and that's why I think we should improve the current design instead of adding support another special case.

Klingseis: Okay, got it. If there were already many requests adapting the guessing, I understand that you want to rethink current design.
My intention was to fix the current approach, because I thought its not that bad except this problem. 
But sure with different files, there might be different problems  where even this implementation might fail.

So your idea is to provide end user a function somehow like guessDelimiter() which returns true or false?

pokoli: The guess delimiter function is currenctly configurable but it sems that there is not so much people using this feature. 
So in order to improve it's adoption I think we need a way to call the standard function from the custom function.

sandinmyjoints:
We are in a similar situation where we'd like to adjust the logic for choosing the best delimiter. OP's proposal, if I'm understanding it right, would help with our case (we're also trying to avoid picking comma when it's not really the delimiter). But in general, I like the idea of making it easier to adjust the guessing logic as @pokoli mentioned.

papaparse.js
Klingseis: By configurable do you mean the arguments you pass to guess delimiter function which are based on config or is there another way to configure it?
Should the default function which PapaParse uses, if there is no custom function, be like the current guess delimiter function?

Klingseis:
Not completed. Was not accepted and discussed to the end.

PR #807, opened 2020-06-24
Add index argument to "transformHeader" config function

furey:
When transforming headers, knowing what column index a header refers to can be extremely useful.

This PR adds the header index as an optional second argument to the `transformHeader` config function.

This PR also corrects the `addHeder` typo (changed to `addHeader`).

Cheers!

🤓👍

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -1189,10 +1189,10 @@ License: MIT
 			if (!_results)
 				return;
 
-			function addHeder(header)
+			function addHeader(header, i)
 			{
 				if (isFunction(_config.transformHeader))
-					header = _config.transformHeader(header);
+					header = _config.transformHeader(header, i);
 
 				_fields.push(header);
 			}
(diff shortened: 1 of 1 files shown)


pokoli:
Hi @furey, 

Thanks for your contribution.

Could you please add a test that ensures the

FabienDeborde:
Hello and thank you very much for this library and this update, I would really really like to start using this functionality but it seems there was no npm update since it was merged into master. 
Is there any way to be able to use this (beside forking the repository) ?

pokoli:
@FabienDeborde as the code is merged into master branch you can build your own papaversion from there.

FabienDeborde:
@pokoli Thank you for your answer. I ended up forking the master branch so I could use npm without having to worry about future updates on the master branch that could contain breaking changes.

By the way, I tried to use the updated transformHeader that should contain the index, but index still returns undefined. 
I am using the code below, if you have time, can you please tell me if I am doing something wrong?
```
Papa.parse(file, {
      skipEmptyLines: true,
      header: true,
      transformHeader: (header, index) => {
        console.log('header', header)
        console.log('index', index)
      }
      complete: onComplete,
      error: onError
})
```

Issue #303, opened 2016-04-08
Suggestion: adding config flag to ignore protoype


shank8:
I am using PapaParse alongside AngularJS and due to the prototypal inheritance that an Angular $resource adds to the JSON, PapaParse is doing a JSON to CSV conversion and adding those prototype properties/methods to the CSV. I added 2 lines of code into the `ObjectKeys` function in order to fix this.  I would be willing to add an implementation of the config object to add this functionality when needed. Let me know if you think this is a valid request and I will fork and add this.

```
function objectKeys(obj)
{
        if (typeof obj !== 'object')
            return [];
        var keys = [];
        for (var key in obj)
            if(obj.hasOwnProperty(key)) {
                keys.push(key);
            }
        return keys;
}
```

pokoli:
@shank8 could you please add some example code on how you are using Paparse with angular. 
Also if you have some code suggestion feel free to open a PR so others can review them.

pokoli:
As disscues on #810 I do not think we should add this configuration option.

PR #810, opened 2020-06-26
added check for property own to ignore prototype chain

ilich-x:
I'm using library (method `unparse` to generate csv) with objects which made from class with static property (using babel). Because of using `for..in` loop in `objectKeys` all methods from prototype chain contained in generated CSV (function methods etc.).
To prevent this i added check for own properties.
This behavior mentioned in  #303

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -376,8 +376,11 @@ License: MIT
 			if (typeof obj !== 'object')
 				return [];
 			var keys = [];
-			for (var key in obj)
-				keys.push(key);
+			for (var key in obj) {
+				if (Object.prototype.hasOwnProperty.call(obj, key)) {
+					keys.push(key);
+				}
+			}
 			return keys;
 		}
 


pokoli:
Hi, 

Thanks for oyur suggestion.

Won't be better to pass directly to paparse a dict with only the propierties to be exported? 
This way we support the current behaviour and the one you need.

ilich-x:
In my case i have a lot of objects to be export in one csv (10000 - 100000 in average) with different count of properties (i need them all), so in current behavior i have to copy each of them. I don't think this is a good way.

Could you please tell why there is a `for..in` loop and why we exactly need full prototype chain, thanks.

pokoli:
If you pass an object you are asking to use it's full prototype chain. 
If you do not want to render all it's full prototype chain just create a new object with the desired properties and pass it to PapaParse.

Note that PapaParse objective is to read/write CSV files but not manipulate data, this is the responsibility of developer using PapaParse. 

I'm closing this issue as there is nothing to do from our side. 
If you have some concers about unprase performance please open a separate issue.

PR #816, opened 2020-07-08
Update docs.html

pschlatt:
For most of us laymen glancing at these docs, the description (or greedy) isn't descriptive enough.  I've seen several unresolved Stack Overflow and other forum posts confused about this.


pokoli:
Merged it. Tjanks!

PR #818, opened 2020-07-23
chunkSize documented

### LINKED ISSUE #673: Add chunkSize to documentation
Hello, while attempting to use this library, I came across an issue.

Perhaps because of the way I was importing it, but I could not get `LocalChunkSize`, or `RemoteChunkSize` to work. (Bug #674)
```
Papa.LocalChunkSize = 1024 * 1024;
Papa.parse(localBlob, {
  chunk: function(...){...}
});
```
Results in the entire file read in a single chunk.

However, after reading the code, this works:
```
Papa.parse(localBlob, {
  chunkSize: 1024 * 1024,
  chunk: function(...){...}
});
```

This is not in the online documentation, and it really should be!

PS:  Everything else is done by passing options, why does `LocalChunkSize` and `RemoteChunkSize` even exist?

akshaybabloo:
closes #673

### DIFF
--- a/docs/docs.html
+++ b/docs/docs.html
@@ -637,7 +637,7 @@ var csv = Papa.unparse({
 									<code>chunkSize</code>
 								</td>
 								<td>
-									Overrides <code>Papa.LocalChunkSize</code> and <code>Papa.RemoteChunkSize</code>.
+									Overrides <code>Papa.LocalChunkSize</code> and <code>Papa.RemoteChunkSize</code>. For more information see <a href="#configurable">configurable</a>.
 								</td>
 							</tr>
 							<tr>
--- a/docs/docs.html
+++ b/docs/docs.html
@@ -444,6 +444,7 @@ var csv = Papa.unparse({
 	downloadRequestBody: undefined,
 	skipEmptyLines: false,
 	chunk: undefined,
+	chunkSize: undefined,
 	fastMode: undefined,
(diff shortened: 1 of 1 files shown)


docs/docs.html
```diff
@@ -631,6 +632,14 @@ <h5 id="config-details">Config Options</h5>
 									A callback function, identical to step, which activates streaming. However, this function is executed after every <i>chunk</i> of the file is loaded and parsed rather than every row. Works only with local and remote files. Do not use both chunk and step callbacks together. For the function signature, see the documentation for the step function.
 								</td>
 							</tr>
+							<tr>
+								<td>
+									<code>chunkSize</code>
+								</td>
+								<td>
+									Overrides <code>Papa.LocalChunkSize</code> and <code>Papa.RemoteChunkSize</code>.
```
pokoli, line 640: I think we should explain what do both values instead of just saing that it overrides the behaviour.

akshaybabloo: Ah I didn't do that because I thought they were already explained at the end of the document. Do you want me to link them to https://www.papaparse.com/docs#configurable or should I explain them here again?

pokoli: Yes please, just add the link.

akshaybabloo: Added 😄

pokoli:
Merge it! Thank you so much!

PR #820, opened 2020-07-24
Update lovers.js

UnnitMetaliya:


pokoli:
Added it. Thanks for your feedback

PR #836, opened 2020-10-08
Added option quoteDataWithSpaces (default - true) to unparse config.

alexander-gush:
If quoteDataWithSpaces set to false, unparse result doesn't quote data with spaces in the start or in the end.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -285,6 +285,9 @@ License: MIT
 		/** whether to prevent outputting cells that can be parsed as formulae by spreadsheet software (Excel and LibreOffice) */
 		var _escapeFormulae = false;
 
+		/** whether to surround every datum that has spaces in the start or in the end with quotes */
+		var _quoteDataWithSpaces = true;
+
 		unpackConfig();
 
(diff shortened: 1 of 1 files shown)


pokoli:
Hi @alexander-gush,

Thanks for your pull request. Could you please explain why this feature is usefull?
In order to accept a PR we require to have a minimal test on our test suite that ensures that your behaviour works as expected. Could you please add one?

alexander-gush:
Thank you!
Some times it may be unexpectedly that quotes appears in output but there are not in input. So I think we should give possibility to control it.
I added test.

pokoli:
We are still missing the doc update (which is in the docs folder of the repository) to include the new option

alexander-gush:
Should I do this?

alexander-gush:
> We are still missing the doc update (which is in the docs folder of the repository) to include the new option

I added this option to the documentation.

pokoli:
Hi @alexander-gush,  

There are some conflicts with the master branch. Could you plesae rebase your work on top on it and fix them? 

Thanks in advance!

PR #842, opened 2020-11-14
Removed unncessary utility fn

jimmywarting:
Saw you used Object.keys elsewhere so I thought it would be fine to use it directly instead


pokoli:
Merged it.

Thanks for removing some old code!

jimmywarting:
Np, mostly did it just to familiarize myself to the codebase a bit.
I'm interested in getting support for async iterator as a lower level api,

Issue #345, opened 2016-09-19
Not breaking lines in files with both \r and \n


burchill:
I don't know if there's an easy way to solve this problem, but when parsing files that vary in their use of what newline character they use (i.e. switching between "\r" and "\n") PapaParse doesn't recognize all these line breaks.  When working with both Windows and Unix based systems, this mish-mash of newline characters can sometimes happen (as it did with me).

I uploaded a small file that demonstrates this for replicability:

https://github.com/burchill/burchill.github.io/blob/master/faulty.csv

pokoli:
@burchill the csv is no longer available, so I can not test it. Could you provide a [demo](http://papaparse.com/demo) input in order to reproduce it or a test scenario?

evanplaice:
The 'easy' solution is to ignore /r linebreaks altogether. But, for some inane reason the CSV output from MS Excel for Mac still uses /r by default; despite the fact that OSX is and has always been *nix based.

I'd recommend preprocessing the CSV and match/replace line endings to be consistent (ie either /n or /r/n).

pokoli:
@evanplaice and how to determine which is the good line delimiter to make the file consistent?

wshostak:
I haven't submitted yet because it does break the test and I have not had time to fix the test.
Right before the comment;
```
// We don't need to compute some of these every time parse() is called,
```
Paste in;
```
// Replace possible EOL to all match
input = input.replace(/[\r\n]+/g, newline);
```
This allows for mixed line breaks.

pokoli:
@wshostak could you create a pull request so we can see what tests are failing?

pckhoi:
Other CSV parsers such as CSV decoder from Go's standard library can handle such files with grace. I don't think this is something that should be ignored by PapaParse.

pckhoi:
@pokoli check out this PR https://github.com/mholt/PapaParse/pull/845 no tests failing.

pokoli:
As commented on the PR I think the input should be modified before parsing to paparse. 
If you now the right delimiter you should replace it before passing it to papaparse.

Issue #826, opened 2020-08-12
Add browser streaming capability


Josh-Weston:
Is there any desire for a pull request to add browser streaming functionality? I have implemented it for an isolated use case:

- I needed to stream data to the browser
- The data is streamed from a database
- I do not know the content-length or byte ranges ahead of time
- The target browser supports **fetch()** and  **ReadableStream API**

Going through the source code I could not find what I was looking for, so I implemented the following in my cloned repo:

```
// Pipes browser streaming capability into Papa Parse
// Useful when content-length is unkown (e.g. streaming from a database)
function BrowserStreamer(config) {
	config = config || {};
	if (!config.chunkSize)
		config.chunkSize = Papa.RemoteChunkSize;
	ChunkStreamer.call(this, config);
	const decoder = new TextDecoder('utf-8');
	this.stream = async function(url)
	{
		try {
			const stream = await fetch(url).then((res) => res.body.getReader());
			this._nextChunk = function() {
				stream.read().then(({ done, value }) => {
					if (done) {
						this._finished = true;
						this.parseChunk('');
						return;
					}
					this.parseChunk(decoder.decode(value));
				});
			};
			this._nextChunk();	// Starts streaming
		} catch (err) {
			this._finished = true;
			this.parseChunk(`Error accessing or parsing data: ${err}`);
			return;
		}
	};
}
BrowserStreamer.prototype = Object.create(ChunkStreamer.prototype);
BrowserStreamer.prototype.constructor = BrowserStreamer;
```

```
// Add another config option
if (_config.browser) {
	streamer = new BrowserStreamer(_config);
```

```
// Accessing the new capability 
const url = 'http://someurl.com';
Papa.parse(url, {
    browser: true,
    delimiter: ',',
    newline: '\n',
    complete: () => {
        console.log('All done!')
    },
    step: (results, parser) => {
        console.log('step results');
        console.log(results.data);
    }
});
```

This was a quick implementation that needs more guard rails and syntax regression, but it should provide enough context. If it's something of value, I can add some rigor and submit as a pull request.

pokoli:
Hi @Josh-Weston 

Yes this adds some value please create a pull request with redefined code so we can review. 
We require to have a minimal tests case to test that the new functionalit works correctly. 

BTW: We have an option to download a file from a URL, maybe the setting will be a complement of download to just enable the sreaming. Could you please check it?

pckhoi:
+1

MichaelSp:
[...]

PR #845, opened 2020-11-28
Add option `forceUniformNewline`

pckhoi:
This will solve this issue https://github.com/mholt/PapaParse/issues/345 with files where newline character vary throughout the file. This solution isn't ideal because it will also modify quoted content. A more permanent solution might be to handle the newline character more dynamically. It seems like a huge effort to me. This solution is good enough for my need.

Please let me know how this look and I will add more tests/docs as needed.


pokoli:
Hi @pckhoi,

THanks for your pull requests. Could you please fix the following issues:

- Do not modify minified file, this is done on the release process
- Could you please document the new parameter in the docs folder? 

Thanks in advance.

pckhoi:
Shouldn't the minified file be in .gitignore then?

pokoli:
Please keep the changes of a pull request minimum and do not include unrelated. You should not modify the gitingore nore the minified file. 

If you think we should ignore the minifed file on git please open a new issue for its discussion.

pokoli:
I'm not sure if we should accept this PR. The reason is clear: 

The objective of PapaParse is just to *parse* but not to modify the original files. 
Is there any issue that prevents you to modify the read data before pasing it to papaparse?

pckhoi:
This is not modifying the file, this is modifying what papaparse output. And this is an option for those who need it, the default behavior is not altered in any way.

I'm streaming the original CSV file to the browser using `chunk` callback and there's no way for me to intercept the stream before it get passed to papaparse. The CSV files are uploaded by our users and as you said, modifying the original CSV files is not an option for us.

If you still want to proceed then I will make the changes you suggested.

pokoli:
Well the problem is that I find it too specific as we only handle the delimiters and I always prevent to have one option for each case. 
From your code I see that you replace all the delimiters to the input variable to the prefered delimiter. The input is what PapaParse gets as input and what your code passes to it so it should be a way to do this on your code without adding anything to the library. 

So I think we must reject this PR. Sorry.

pckhoi:
I could work around this if papaparse accept [ReadableStream](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream) as input.

pokoli:
Then it's a mather of implementing #826 which will solve most of the feature requests.

PR #852, opened 2021-01-11
Drop redundant getNextUnquotedDelimiter function

### LINKED ISSUE #782: getNextUnquotedDelimiter function seems to be useless
Hello.
Maybe I'm missing something really obvious but it seems that `getNextUnquotedDelimiter` is useless now.

It gets called from 2 places - from the same function on [line 1776](https://github.com/mholt/PapaParse/blob/master/papaparse.js#L1776) and from the outside on [line 1630](https://github.com/mholt/PapaParse/blob/master/papaparse.js#L1630)

Let's check this outside call:
```javascript
// we check, if we have quotes, because delimiter char may be part of field enclosed in quotes
if (quoteSearch > nextDelim) {
    // we have quotes, so we try to find the next delimiter not enclosed in quotes and also next starting quote char
    var nextDelimObj = getNextUnquotedDelimiter(nextDelim, quoteSearch, nextNewline);
```

Now let's check the function definition:
```javascript
/** Gets the delimiter character, which is not inside the quoted field */
function getNextUnquotedDelimiter(nextDelim, quoteSearch, newLine) {
    var result = {
        nextDelim: undefined,
        quoteSearch: undefined
    };
    // get the next closing quote character
    var nextQuoteSearch = input.indexOf(quoteChar, quoteSearch + 1);

    // if next delimiter is part of a field enclosed in quotes
    if (nextDelim > quoteSearch && nextDelim < nextQuoteSearch && (nextQuoteSearch < newLine || newLine === -1)) {
        // some code skipped here
    } else {
        result = {
            nextDelim: nextDelim,
            quoteSearch: quoteSearch
        };
    }

    return result;
}
```
Turns out this `if` condition in function will never happen because `nextDelim` will always be less than `quoteSearch` in this outside function call and function just returns passed arguments.

FarSeeing:
Closes #782

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -1625,27 +1625,11 @@ License: MIT
 				// Next delimiter comes before next newline, so we've reached end of field
 				if (nextDelim !== -1 && (nextDelim < nextNewline || nextNewline === -1))
 				{
-					// we check, if we have quotes, because delimiter char may be part of field enclosed in quotes
-					if (quoteSearch > nextDelim) {
-						// we have quotes, so we try to find the next delimiter not enclosed in quotes and also next starting quote char
-						var nextDelimObj = getNextUnquotedDelimiter(nextDelim, quoteSearch, nextNewline);
-
-						// if we have next delimiter char which is not enclosed in quotes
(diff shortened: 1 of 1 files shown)


pokoli:
Merged in. Thanks!

PR #863, opened 2021-03-22
Change float regex

ConnorSmith-pf:
Mathematica uses uppercase and lowercase. Euler's number therefore can be uppercase or lowercase, however only lowercase was checked for in the FLOAT RegEx.


pokoli:
Hi @ConnorSmith-pf 

Thanks for creating a pull request. 
Could you plese remove the commit that updates the minified file? This is automatically managed by our release process.

ConnorSmith-pf:
Hi @pokoli 

Apologies, I've reverted that commit!

Issue #612, opened 2019-01-06
[Idea/Suggestion] Improved support for "header" lines in input files


Eric24:
A rather common CSV format found in things like web server logs is to have one or more "header" lines at the beginning of the file, like this:
```
#Version: 1.0
#Fields: date time x-edge-location sc-bytes c-ip cs-method cs(Host) cs-uri-stem sc-status cs(Referer) cs(User-Agent) cs-uri-query cs(Cookie) x-edge-result-type x-edge-request-id x-host-header cs-protocol cs-bytes time-taken x-forwarded-for ssl-protocol ssl-cipher x-edge-response-result-type cs-protocol-version fle-status fle-encrypted-fields
2019-01-04	11:37:39	ORD50-C1	1030	0.0.0.0	GET	xxxxxxxxxx.domain.net	/	200	-	Mozilla/5.0%2520(Windows%2520NT%252010.0;%2520Win64;%2520x64)%2520AppleWebKit/537.36%2520(KHTML,%2520like%2520Gecko)%2520Chrome/71.0.3578.98%2520Safari/537.36	-	-	Test	qiW8xjO4CoRnsrW843xPY4fE_sKPZp7YYgpeskJhd3N9ZkArsmUvrw==	host.domain.com	https	283	0.188	-	TLSv1.2	ECDHE-RSA-AES128-GCM-SHA256	Miss	HTTP/2.0	-	-
```

Obviously this library can't be expected to deal with every such header format, but I think this could be solved fairly simply, by adding two props to `options`: a prefix to identify such header lines; and a function (maybe called something like 'parseHeaders'). The idea being to allow a custom function to have a look at the first n lines of the file, until the function figures out the headers and returns them.

When the library starts processing a file, its in "seeking header information" state and not parsing data. Every line that starts with the specified prefix is passed to the parseHeaders function (the entire line). That function can either return a falsey value or a hash or array that contains the headers. Once the headers are returned, the library is now in "seeking data" state (but it will continue passing any line identified by the prefix to this function, in case there are other such headers after the one that contains the column headers).

In any case, the general idea of getting the column headers from the file is already there, but it's currently limited to only looking for headers on the first non-comment/non-blank row and then they must follow the same format as the data, with regard to delimiters, etc. This concept allows extendable support for more complex header formats, including cases where the column headers may not even be itemized in the file at all, but are instead identified by some sort of constant, etc.

[...]

PR #898, opened 2021-11-03
Implement https://github.com/mholt/PapaParse/issues/612

theLAZYmd:
- Modify transformHeader to take 3 new parameters:
-- The whole headers row as an array (arr)
-- The previous value of the header in that column (acc)
-- The row number (j)
- Applies the transformHeader function over each row in data up to your headerLines value
- Creates a config.headerLines parameter (default 1)

### DIFF
--- a/docs/docs.html
+++ b/docs/docs.html
@@ -510,12 +510,16 @@ var csv = Papa.unparse({
 									<code>transformHeader</code>
 								</td>
 								<td>
-									A function to apply on each header. Requires <code>header</code> to be <code>true</code>. The function receives the header as its first argument and the index (the column number) as second. When used in conjunction with config.headerLines, the function receives a third and fourth parameter which are:
+									A function to apply on each header. Requires <code>header</code> to be <code>true</code>. The arguments are:
 									<ul>
+										<li><b>header:</b> the current header text, as parsed from the appropriate cell in the specified header row</li>
+										<li><b>index:</b> the column number of the header text (0 for first column)</li>
 										<li><b>arr:</b> the string array of data for that row</li>
 										<li><b>acc:</b> the existing header for that column based on previous lines iteration ('' if no previous lines)</li>
 										<li><b>j:</b> the rowNumber (0 for the first line)</li>
 									</ul>
+									Only available starting with version 5.0.
+									The latter three parameters (arr, acc, j) are only available when used in conjunction with config.headerLines and from version 5.4+.
 								</td>
 							</tr>
 							<tr>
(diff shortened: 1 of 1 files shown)


pokoli:
This is related to #612

pokoli:
Please do not update the minified file, this will be done by our release process. 
Could you please update the review to revert such changes.

docs/docs.html
```diff
@@ -510,8 +510,20 @@ <h5 id="config-details">Config Options</h5>
 									<code>transformHeader</code>
 								</td>
 								<td>
-									A function to apply on each header. Requires <code>header</code> to be <code>true</code>. The function receives the header as its first argument and the index as second.<br>
-									Only available starting with version 5.0.
```
pokoli, line 514: I think we should keep that this is only availalbe on the version 5.0

I also plan to do a minor version for the parameter. So the new parameter structure should be only available on 5.4+ versions. That should be noted too.

papaparse.js
```diff
@@ -1180,24 +1180,26 @@ License: MIT
...
+				{
+					// A function which takes two arguments (header, i) where j is set by the fact it is called in this loop. It then calls addHeader() with all three arguments
+					_results.data[j].forEach(addHeader.bind(null, j));
+				}
+				_results.data.splice(0, headerLines);
 			}
 			// if _results.data[0] is not an array, we are in a step where _results.data is the row.
 			else
-				_results.data.forEach(addHeader);
+				_results.data.forEach(addHeader.bind(null, 0));
```
pokoli, line 1202: Why is bind required? Could we just ignore make the arguments at the end and optional?

pokoli (changes requested):

docs/docs.html
theLAZYmd: Resolved with https://github.com/mholt/PapaParse/pull/898/commits/ac6f19168cc742ca12ef7ee3c2dcae607e9301a3

theLAZYmd:
> Please do not update the minified file, this will be done by our release process. Could you please update the review to revert such changes.

Resolved with https://github.com/mholt/PapaParse/pull/898/commits/417982cda9eeb8120106045cbd1f88bb12bc3adc

papaparse.js
theLAZYmd: Bind is required because we're trying to parse a function into the .forEach() method, and unfortunately we have
"no-loop-func": "error"
and ecma5 versioning in our eslint config, which I think means no arrow functions. I assume this is for reasons of universalisable compatibility and speed. As a result, this is the only way to pass the loop iterator into the function.

theLAZYmd:
@pokoli and else, is there any update on this?

pokoli:
We need a bug fix release before merging this. 
I will take care of that!

PR #899, opened 2021-11-07
'papaparse-min' sync workflow script

tjohnston-softdev:
@pokoli - I was able to put together a GitHub Actions workflow script that should be able to automatically sync to the 'papaparse-min' library whenever a new version is released. It is triggered whenever a new tag is created. This is mainly done through 'Releases'. I could not get around to testing whether it triggers when adding a tag through Git itself. You will have to set your own Git credentials, GitHub access token, and NPM access token. 

This is by no means perfect but it seems to get the job done with basic testing. Let me know any feedback.


pokoli:
Hi @tjohnston-softdev , 

Thanks for your proposal, it looks interesenting but I have some concers because we normally just create a release for each minor versions but for bug fixes. 

AFAIU this will require to manually publish version to npm for bug fixes. 

It is possible to make the script work just when a new tag is created on git (no mather if a Releases exists on GH). I'm used to tagging when relasing a new version so this will make it work. 

If not probably not a big issue as I'm used to build the minified version whenever a new version is released as I'm currently doing the process manually in my laptop. But having it integrated in the CI will allow anyone which push rights to the repo to push new versions.

PR #919, opened 2022-02-24
fix: columns config works with input type object

### LINKED ISSUE #918: 'columns' unparse config not working when first object doesn't have all properties
When using the `columns` unparse config option together with a data array where the first object does not have all properties that are listed in the `columns` array, then not all columns of the `columns` array are used even if later objects have these values.

E.g.
```javascript
const columns= ["First", "Second", "Third"];
const rows= [
  { First: "1-1"},
  {
    First: "2-1",
    Second: "2-2",
    "Third": "2-3",
  },
];
const result = Papa.unparse(
  { data: rows},
  {
    header: true,
    columns: columns,
  }
);
console.log(result)
```
The `result` is 
```
First
1-1
2-1
```

instead of the expected

```
First,Second,Third
1-1,,
2-1,2-2,2-3
```

TheSlimvReal:
Closes #918

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -306,9 +306,10 @@ License: MIT
 
 			if (Array.isArray(_input.data))
 			{
-				if (!_input.fields)
-					_input.fields =  _input.meta && _input.meta.fields;
-
+				if (!_input.fields) {
+					_input.fields = _input.meta && _input.meta.fields || _columns;
+					console.log('input', _input.fields);
+				}
 				if (!_input.fields)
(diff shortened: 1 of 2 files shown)


pokoli:
Hi @TheSlimvReal, 

We have introduced some fixes on our CI on the master branch, before merging this PR i will like to run the test suite with the latest configuration. Could you please merge the master branch on your changes?

TheSlimvReal:
@pokoli done

PR #925, opened 2022-03-12
fix: set empty string to config.quotechar when it's value is null

### LINKED ISSUE #914: [Cannot read properties of null (reading 'replace')] error when quoteChar = null
Hi!
I use papa parse for a long time, and that is the first problem I found throughout my experience with it.
It's not very important but it took me some time to understand the source of the issue, 
so even if it's not fixed maybe this github issue will help someone in the future.

### The problem
When the `quoteChar` is set to `null`, and you have `'null'` string in the parsed file - an unexpected exception appears and parsing process terminates without calling any of the (error/complete) callbacks.

### Example
1. Open papa parse [demo page](https://www.papaparse.com/demo) and open console.
2. Execute the following code: 
```
Papa.parse(
    new File(['Franziska;Knull'], 'some-file.csv', {type: 'text/csv', lastModified: 1572878339451}), 
    {delimiter: ';', quoteChar: null, error: () => console.log("error"), complete: () => console.log("complete")},
)
```
3. Observe the error in console:

image

### Workaround
When you do not need any `quoteChar` just set it to `undefined` or `''` (empty string), then it works as expected.

Thanks!

chafi9-code:
Closes #914

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -1403,8 +1403,7 @@ License: MIT
 		var preview = config.preview;
 		var fastMode = config.fastMode;
 		var quoteChar;
-		/** Allows for no quoteChar by setting quoteChar to undefined in config */
-		if (config.quoteChar === undefined) {
+		if (config.quoteChar === undefined || config.quoteChar === null) {
 			quoteChar = '"';
 		} else {
 			quoteChar = config.quoteChar;


chafi9-code:
Hi @mholt can you review my PR please ?
Thank's

papaparse.js
```diff
@@ -1403,8 +1403,8 @@ License: MIT
 		var preview = config.preview;
 		var fastMode = config.fastMode;
 		var quoteChar;
-		/** Allows for no quoteChar by setting quoteChar to undefined in config */
-		if (config.quoteChar === undefined) {
+		/** Allows for no quoteChar by setting quoteChar to undefined or null in config */
```
mholt, line 1406: I think this is obvious, so the comment is not that useful.

```suggestion
```

mholt (approved):
LGTM (except the comment is unnecessary) -- thanks for the patch!

papaparse.js
chafi9-code: effectively, i agree with you.

pokoli:
Hi @chafi9-code, 

It seems your changes are breaking our test suite. Could you please have a look?

chafi9-code:
hi @pokoli.
it's done, i rebased  it on the master branch.

chafi9-code:
hi @pokoli.
it's done, i rebased  it on the master branch.

pokoli:
Hi @chafi9-code we need to fix the tests before we can merge it.
Please have a look at [them](https://github.com/mholt/PapaParse/runs/5535406278?check_suite_focus=true)

chafi9-code:
i did a mistake in the second part of the condition and after checking i fixed it. thinks for the tests

pokoli:
I merged your fix, thank you so much!!

PR #937, opened 2022-06-08
[GH-788] Made the check to determine whether the current web-worker is owned by PapaParse more reliable

### LINKED ISSUE #788: Calling Papa.parse in worker with _config.worker = false calls postMessage
I am calling `Papa.parse` in my own worker and in `_config.worker` is set false. 
When process completes it calls `postMessage` still that triggers a message to my own worker.
Shouldn't [this](https://github.com/mholt/PapaParse/blob/master/papaparse.js#L529) 
```
if (IS_PAPA_WORKER)
{
	global.postMessage({
		results: results,
		workerId: Papa.WORKER_ID,
	});
}
```
Should also check if `config.worker` is set to `true` then only call `global.postMessage`.
Was this done intentionally?

flakey-bit:
This fixes #788

Previously, the assumption was if we're in a web-worker AND the worker was loaded from a blob, the worker *must* be ours.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -49,11 +49,12 @@ License: MIT
 	function getWorkerBlob() {
 		var URL = global.URL || global.webkitURL || null;
 		var code = moduleFactory.toString();
-		return Papa.BLOB_URL || (Papa.BLOB_URL = URL.createObjectURL(new Blob(['(', code, ')();'], {type: 'text/javascript'})));
+		return Papa.BLOB_URL || (Papa.BLOB_URL = URL.createObjectURL(new Blob(["var global = (function() { if (typeof self !== 'undefined') { return self; } if (typeof window !== 'undefined') { return window; } if (typeof global !== 'undefined') { return global; } return {}; })(); global.IS_PAPA_WORKER=true; ", '(', code, ')();'], {type: 'text/javascript'})));
 	}
 
 	var IS_WORKER = !global.document && !!global.postMessage,
-		IS_PAPA_WORKER = IS_WORKER && /blob:/i.test((global.location || {}).protocol);
+		IS_PAPA_WORKER = global.IS_PAPA_WORKER || false;
+
 	var workers = {}, workerIdCounter = 0;
 
 	var Papa = {};
--- a/tests/.eslintrc.js
+++ b/tests/.eslintrc.js
@@ -1,5 +1,8 @@
 module.exports = {
 	"extends": ["../.eslintrc.js"],
+	"parserOptions": {
+		"ecmaVersion": 8
+	},
 	"env": {
 		"mocha": true
 	},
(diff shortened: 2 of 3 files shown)


pokoli:
Hi @flakey-bit, 

Thanks for working on this. I see there is an issue on your patch that match our test suite broken. 
Could you have a look in order to fix it? TIA

flakey-bit:
> Hi @flakey-bit,
> 
> Thanks for working on this. I see there is an issue on your patch that match our test suite broken. Could you have a look in order to fix it? TIA

@pokoli 

Can you link me to the test-suite build failure?

NB: I found there was an issue with my implementation so I've had another go at fixing it 🤞 - I've force-pushed to my branch, which should have updated the PR.

flakey-bit:
@pokoli 

I've added a test case to cover this scenario. Please take a look

pokoli:
Hi @flakey-bit, 

Test failed on node12 which is unsuported by now. 
I removed the support of node 12 on master branch, could you please rebase it? 
If test passes I will merge it.

flakey-bit:
@pokoli rebase is done

pokoli:
@flakey-bit Tests are still broken, could you please have a look?

You can run them locally with: `npm test`

flakey-bit:
I'll take a look tomorrow. Sorry for taking up your time!

flakey-bit:
@pokoli have force-pushed again.

`npm test` is clean now (as is running `npm run test-node` and `npm run test-mocha-headless-chrome` separately.

Note I've included an additional to commit to allow the tests to use latest ECMAScript syntax (does not affect the targeting of the main library).

flakey-bit:
@Silic0nS0ldier can you also take a look at this? The approach will likely have implications for the move to ES6 modules.

In particular, see the test case that this PR adds.

pokoli:
@flakey-bit why is required to add the latest ECMAScript syntax? Does the test work without that?

flakey-bit:
@pokoli the test is currently written using `async` keyword, arrow functions, template strings etc.

Could rewrite the test not to use such features, but would probably make it harder to understand.

Silic0nS0ldier:
@flakey-bit I've lost quite a bit of context since opening #875, but I don't see any reason for this to cause problems for ES6.

Not sure how relevant it will be here, but I did some work to solve the same problem with NodeJS worker threads some time ago. https://github.com/developit/web-worker/blob/9ee077b6c5df0acaf5d83d89e2087bd3a3b82b97/src/node.js#L138-L142

flakey-bit:
@pokoli any chance of getting a new release out to NPM?

flakey-bit:
@pokoli it would be awesome to get a release out if it's not too much trouble

I know it's a bit selfish of me, but part of the reason I picked up #788 was to allow us to integrate PapaParse into our application without resorting to [patch-package](https://www.npmjs.com/package/patch-package) 🙂

PR #950, opened 2022-08-15
Add poulos method to Papa.

LukeGeneva:
I've used this library on a few different occasions and have always thought this would be a funny little Easter-egg-like addition, since it reads like "Papadopoulos."

Anyway, whether it gets merged or not, hopefully it was good for a laugh and I haven't wasted your time.

All the best!

Luke


pokoli:
Thanks for your funny proposal but I think we will keep it serious and don't merge it.

Issue #761, opened 2020-01-30
Error using transformHeader with header: true and worker: true


bleuf1sh:
The error was about not being able to clone in postMessage.

I was working with a remote csv in a Chrome Web Front End.

pokoli:
could you post the full tracebak?

bleuf1sh:
I've provided the example function as well and removed private details.

```
Uncaught (in promise) DOMException: Failed to execute 'postMessage' on 'Worker': (h) => {
                    return h.trim();
                } could not be cloned.
    at Object.parse (chrome-extension://asfdasdf/background/background.js:35623:967)
    at chrome-extension://asfdasdf/background/background.js:46081:18
    at new Promise (<anonymous>)
    at Function.loadCsvRules (chrome-extension://asfdasdf/background/background.js:46079:16)
```

```

private static async loadCsvRules(csvUrl: string): Promise<any[]> {
    return new Promise( (resolve, reject) => {
      const datas = [];
      Papa.parse(csvUrl, {
        download: true,
        header: true,
        worker: true,
        transformHeader: (h) => {
          return h.trim();
        },
        step: (row) => {
          datas.push(row.data);
        },
        complete: () => {
          resolve(datas);
        }
      });
    });
  }

```

abodnar:
I'm having this same problem as well. Any suggestions on how to resolve this?

steveryan:
Having the same issue over here. Would love to find a workaround for this.

ux-engineer:
Same here...

jchan18:
Is there any update with this issue? I'm seeing this problem too

mikeappell:
👍 on this, it seems like `transformHeader` is incompatible with `worker: true` which really stinks. Even
```transformHeader: header => header```
seems like it fails; assuming the function is being passed to the worker and simply cannot be serialized.

Just so as not to poo all over an amazing project: PapaParse is really excellent and has proven utterly invaluable, so thanks for such an awesome tool 🎊

jaymathew:
I'm a new PapaParse user, and I wish this were fixed or documented. It would have saved me some time.  The last message was over a year ago, and I hate to beat a dead horse, but is there any update on this issue?

joaquinricci:
Same thing happen to me but with `transform` and not `transformHeader`.

m1m1s1ku:
Same issue in 2025

Tremorok:
same issue in 2026

Issue #623, opened 2019-01-28
Failed to execute 'postMessage' on 'Worker' when worker:true and using transformHeader function


UlTriX:
Using Papa v5.0.0-beta.0 trying to use transformHeader config as follow:

`Papa.parse(file, {   
                header: true,
                preview: 1000,
                worker: true,
                transformHeader: function (header) {
                    return header;
                }
            });` 

and getting error:

`Error: Failed to execute 'postMessage' on 'Worker': function (header) {
                    return header;
                } could not be cloned.`

nbrysondev:
I have the same issue for the transform callback. Is this feature fully implemented?

fabricedesre:
Functions are not transferable so in worker mode a different strategy should be used. I can think of 2 options:
- postMessage to the main thread to run the transformation functions (likely slow).
- define the transform functions in their own file and import it in the worker with importScripts('...');

rbrundritt:
It is possible to pass a simple function to a web worker, although a bit hackish. You would convert the function to a string then use eval in the work to re-inflate it. For example:

From the UI thread. 

```
var headerTransform = function(h) { return h.trim(); };

worker.postMessage({
      transformHeader: headerTransform.toString()
});
```
In the worker:

```
self.onmessage = function (e) {
    if (e.data.transformHeader) {
        eval('var transformHeader= ' + e.data.transformHeader);

       //Use the transform header.
        transformHeader(e.data.data);
    }

    //...
};
```

pokoli:
@rbrundritt wow, this is really tricky but I do not think there is another workarround.

Do you think we can document it and thus close this issue?

fabricedesre:
That won't work if you use a CSP that prevents eval() to protect against XSS though...

AlayVora:
Going through same issue. I do not see any solution or documentation related to this issue.

ac-interlace:
@mholt is this fix/workaround for this released? If not, which version can we expect it in? Thanks

mholt:
Sorry, I haven't actually been maintaining this library for several years now. @pokoli is your person! :)

0biWanKenobi:
ping

pokoli:
@0biWanKenobi do you have any proposal to solve the issue?

Fanna1119:
any update on this?

KoalaBear84:
I'm getting the same for `beforeFirstChunk` and `chunk`. No idea how to solve it.

```javascript
[...]

PR #953, opened 2022-09-05
Fix transformHeader can't work with worker (#623) (#761)

imReker:
When worker set to true, convert transformHeader function to string, so the config could be cloned to worker and convert back in worker thread by `eval`.
So, when using Content Security Policy, this workaround will still fail, consider to use vkThread instead.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -201,6 +201,7 @@ License: MIT
 
 		_config.transform = isFunction(_config.transform) ? _config.transform : false;
 
+
 		if (_config.worker && Papa.WORKERS_SUPPORTED)
 		{
 			var w = newWorker();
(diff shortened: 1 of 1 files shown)


pokoli:
I'm a little bit worried about the eval function as string specially as anyone can manipulate the string definition and then inject malicious code on the worker. 

This raises a BIG WARNING on my mind. 

Can we avoid using eval?

imReker:
> I'm a little bit worried about the eval function as string specially as anyone can manipulate the string definition and then inject malicious code on the worker.
> 
> This raises a BIG WARNING on my mind.
> 
> Can we avoid using eval?

Yes, there are potential risks. But the `transformHeader` usually doesn't come from dynamic user input, and the eval is running in the worker, so only global data and codes could be executed. Maybe a big notice in the document is needed . 
And, as I mentioned, to totally avoid `eval`, consider using vkThread(https://github.com/vkiryukhin/vkthread), but I think it's a bit heavy for this problem.

dbaranoff:
[here's the way](https://gkhanko.medium.com/how-to-run-a-javascript-function-in-a-separate-thread-via-web-workers-85d8e36f5847) to do it without `eval`

Issue #129, opened 2014-12-04
Duplicate header names cause parsing to overwrite data


jannah:
I am testing a system I am building using PapaParse. One scenario I am looking it is if more than one column shared the same header. What I am seeing is that only the last column would be included in the result.
http://jannah.github.io/MetaReader
Is there a way to rename the duplicate columns?

Overdrivr:
That's not so rare, happened to me yesterday ;)
For data pipelines, this is very annoying because it silently fails (in my case, the column in question was an "id" one, breaking all relationships in my data model).
At the bare minimum, would it be acceptable to have an error returned because fundamentally the behavior is undefined ?

fortydegrees:
Likewise had this issue. I crudely applied the fix in @jannah 's fiddle to the parse function in papaparse.js

Line 1460, old:

```
if (fastMode || (fastMode !== false && input.indexOf(quoteChar) === -1))
			{
				var rows = input.split(newline);
				for (var i = 0; i < rows.length; i++)
				{
					row = rows[i];
					cursor += row.length;
					if (i !== rows.length - 1)
						cursor += newline.length;
					else if (ignoreLastRow)
						return returnable();
					if (comments && row.substring(0, commentsLen) === comments)
						continue;
					if (stepIsFunction)
					{
						data = [];
						pushRow(row.split(delim));
						doStep();
						if (aborted)
							return returnable();
					}
					else
						pushRow(row.split(delim));
												
					if (preview && i >= preview)
					{
						data = data.slice(0, preview);
						return returnable(true);
					}
				}
				return returnable();
			}
```

Updated to append "_" to duplicate fields:

```
if (fastMode || (fastMode !== false && input.indexOf(quoteChar) === -1))
			{
				var rows = input.split(newline);
				for (var i = 0; i < rows.length; i++)
				{
					row = rows[i];
					cursor += row.length;
					if (i !== rows.length - 1)
						cursor += newline.length;
					else if (ignoreLastRow)
						return returnable();
					if (comments && row.substring(0, commentsLen) === comments)
						continue;
					if (stepIsFunction)
					{
						data = [];
						pushRow(row.split(delim));
						doStep();
						if (aborted)
							return returnable();
					}
					else

						if (i == 0){
							let separator = '_';
							let headers = row.split(delim) //headers array
							let header_map = []; //map of headers after resolving duplicates
							let header_count = {}; //count the headers
							//process headers
							for (var i in headers) {
								var header = headers[i]
[...]

PR #956, opened 2022-10-24
Added support for renaming duplicate headers

fortydegrees:
Added code to automatically rename headers to solve the issue in #129 

I tried to add a renameDuplicateHeaders config option but left it out as I am doing the processing in the Parse.parse function where I don't have the ability to read the config. I'm not sure it fits the styling of the codebase to modify the input outside of that function so have left it out.


fortydegrees:
Thanks for the quick feedback & sorry for the misunderstanding re: test cases.

I've added a test case, removed the useless comments, and added a conditional based on whether the header is present. 

You were right - the renaming code could be refactored. I originally just took the code from that issue thread which worked, but rewrote it in a simpler way which produces the same result.

papaparse.js
```diff
@@ -1457,6 +1457,34 @@ License: MIT
...
+			// Rename headers if there are duplicates
+			if (config.header)
+			{
+				let firstLine = input.split(newline)[0]
+				let separator = '_';
+				let headers = firstLine.split(delim) 
+				let header_map = []; 
+				let header_count = {};
+
+				for (var i in headers){
```
pokoli, line 1469: I guess it will be easier to read if you write: 

var header = headers[i]; 

and then use `header` instead of `headers[i]` in all of the loop code.

papaparse.js
```diff
@@ -1457,6 +1457,34 @@ License: MIT
 			if (!input)
 				return returnable();
 
+			// Rename headers if there are duplicates
+			if (config.header)
+			{
+				let firstLine = input.split(newline)[0]
+				let separator = '_';
+				let headers = firstLine.split(delim)
```
pokoli, line 1465: Could you add `headers`variable after `firstLine`? As they are linked it make sense to define one after the other.

papaparse.js
```diff
@@ -1457,6 +1457,34 @@ License: MIT
...
+						header_map.push(headers[i])
+					}
+					else{
+						if (!(headers[i] in header_count)){
+							header_count[headers[i]] = 1
+						}
+						else{
+							header_count[headers[i]] +=1
+						}
+						header_map.push(headers[i]+separator+header_count[headers[i]])
```
pokoli, line 1480: I will prefer to have a single line that pushes to `header_map`.
So you just need to have a variable name `header_name` that is initialized with the same value as `header`

Then if `header` is already in `header_map+  you just save the count and update the variable name.

tests/test-cases.js
```diff
@@ -585,6 +585,15 @@ var CORE_PARSER_TESTS = [
 			data: [['a', 'b', 'c'], ['']],
 			errors: []
 		}
+	},
+	{
+		description: "Duplicate header names",
```
pokoli, line 590: Could you add a test with also transformHeader function? As we are "manipulating" the headers I think we need to ensure that both functionalities work together.

pokoli (changes requested):
Thanks for your work on this. Everything looks nice, I just added some comments to improve code redability and ensure everything works as expected. Could you please review them?

fortydegrees:
Have addressed all comments re: code-cleanup.

Good call on the transformHeader() test. I added it and It originally failed. I added a check for the function when initialising the header in my code. Hopefully this is satisfactory.

I also added an extra test in the case of an existing header matching an intended substitution. An unlikely situation but thought it'd be wise to cover it.

tests/test-cases.js
```diff
@@ -585,7 +585,34 @@ var CORE_PARSER_TESTS = [
...
+		description: "Simple duplicate header names",
+		input: 'A,A,A,A\n1,2,3,4',
+		config: { header: true },
+		expected: {
+			data: [['A', 'A_1', 'A_2', 'A_3'], ['1', '2', '3', '4']],
+			errors: []
+		}
+	},
+	{
+		description: "Duplicate header names where an original header matches intended substitution",
```
pokoli, line 599: I do not have a clear way that this is the best way to do. Probably we should respect the original names, but this makes code more complex.

As this is a corner case I think the best is to remove such test, and let the user decide what to do by using the transformFunction confiuguration.

Otherwise if we keep the test we are agreeing that this is the expected behaviour and I do not think is the right way.

papaparse.js
```diff
@@ -1457,6 +1457,37 @@ License: MIT
...
+				let header_map = []; 
+				let header_count = {};
+
+				for (var i in headers){
+					var header = headers[i]
+					if (isFunction(config.transformHeader))
+						header = config.transformHeader(header, i);
+					var header_name = header
+					if (header_map.includes(header_name)){
+						if (!(header in header_count)){
```
pokoli, line 1475: I think its simpler to write as:

count = header_count[header_name] || '1';
header_name = header+separator+count;
header_count[header_name] = count +1;

Removeing the if else lines.

pokoli (changes requested):
Everythinkg looks nice now. I leaved some minor issues that should be addressed.

Could you fix them? I think we will be able to merge it once they are addressed. 

Thanks for your time on this.

papaparse.js
fortydegrees: I struggled to make it work without an if statement. The problem being that it would still add the suffix at the end. The best I got was:

`let count = header_count[header] || 0
if (count > 0) header_name = header+separator+count;
header_count[header] = count+1`

Have committed this for you consideration if you prefer.

tests/test-cases.js
fortydegrees: I agree. Have removed the test.

papaparse.js
```diff
@@ -1457,6 +1457,32 @@ License: MIT
 			if (!input)
 				return returnable();
 
+			// Rename headers if there are duplicates
+			if (config.header)
+			{
+				let firstLine = input.split(newline)[0]
```
pokoli, line 1463: please add ; as ending char on all lines to follow the same style.

papaparse.js
```diff
@@ -1457,6 +1457,32 @@ License: MIT
...
+						header = config.transformHeader(header, i);
+					var header_name = header
+
+					let count = header_count[header] || 0
+					if (count > 0) header_name = header+separator+count;
+					header_count[header] = count+1
+
+					header_map.push(header_name)
+				}
+				var edited_input = input.split(newline)
```
pokoli, line 1481: We should avoid spliting the input when the computed header is the same as the previous one. 
This will avoid some performance degradation on bigger files without duplicated headers.

So I think we should: 

1. Get the first line (you already have it as `headers` variable)
2. Compute the new header: `header_map.join(delim)`
3. If both values are different alter the input (current code). Otherwise do nothing.

papaparse.js
```diff
@@ -1457,6 +1457,32 @@ License: MIT
...
+				let header_count = {};
+
+				for (var i in headers){
+					var header = headers[i]
+					if (isFunction(config.transformHeader))
+						header = config.transformHeader(header, i);
+					var header_name = header
+
+					let count = header_count[header] || 0
+					if (count > 0) header_name = header+separator+count;
```
pokoli, line 1476: ok, that looks nicer but keep the if statement in separate lines: 

```
if (count > 0) {
    header_name = header+separator+count;
}
```

pokoli (changes requested):

pokoli (approved):

pokoli:
@fortydegrees linting is failing, could you have a look? 

https://github.com/mholt/PapaParse/actions/runs/3463298598/jobs/5783355405

papaparse.js
fortydegrees: Great catch. I've updated this now. It will still fire if transformHeaders is called. This can be fixed by setting a flag i.e. `edited = true` in:

```
if (count > 0) {
	headerName = header + separator + count;
}
```

and then checking for that flag instead of comparing the two arrays. Otherwise it works fine. Let me know if you'd rather this solution.

fortydegrees:
> @fortydegrees linting is failing, could you have a look?
> 
> https://github.com/mholt/PapaParse/actions/runs/3463298598/jobs/5783355405

Should be fixed now. Thank you for your excellent help in cleaning everything up!

pokoli:
@fortydegrees I've merged your work. Thanks you so much!

sethmcleod:
Thank you for resolving this @fortydegrees!

@pokoli Would you mind doing another minor version bump so we can use these changes in production?

PR #971, opened 2023-02-14
Don't call postmessage when config.worker == false #970

### LINKED ISSUE #970: Calling Papa.parse in worker with _config.worker = false calls postMessage
This bug is well described in [stackoverflow](https://stackoverflow.com/questions/61513192/calling-papa-parse-in-worker-with-config-worker-false-calls-postmessage):

I am calling Papa.parse in my own worker and in _config.worker is set false. When process completes it calls postMessage still that triggers a message to my own worker. Shouldn't [this](https://github.com/mholt/PapaParse/blob/master/papaparse.js#L529)

```
if (IS_PAPA_WORKER)
{
    global.postMessage({
        results: results,
        workerId: Papa.WORKER_ID,
    });
}
```

Should also check if config.worker is set to true then only call global.postMessage I don't want to send results to main thread any idea how can i achieve it ?

The accepted answer is a hack that no longer works, because self.document is not writable

andreaskundig:
Fixes #970
I changed one line that fixes the problem in my own project.

I'm on windows so I couldn't run the tests, not even in wsl.


andreaskundig:
I'm calling papaparse from a worker, so I don't want it to run with its own worker, so I set its config.worker = false. I want to deploy my project to a cdn, to avoid the [cross-origin problem](https://stackoverflow.com/questions/21913673/execute-web-worker-from-different-origin), the workers need to be inlined.

Webpack can do the inlining, which ends up as one big string where the code of papaparse is  concatenated with the code of my worker. My worker builds a message that it sends to its caller with postmessage. However, papaparse mistakenly [also calls postmessage](https://stackoverflow.com/questions/61513192/calling-papa-parse-in-worker-with-config-worker-false-calls-postmessage) #970 so my caller receives two messages: the output of my worker, and the output of papaparse's worker. My fix prevents the mistaken call to postmessage.

pokoli:
Cross origin is what is described in the [standard](https://html.spec.whatwg.org/multipage/workers.html#infrastructure-2). I do not think we should support behaviour that goes agains the standard.

andreaskundig:
What I'm trying to do does not go against the standard. The standard explicitely mentions it:

> Any [same-origin](https://html.spec.whatwg.org/multipage/browsers.html#same-origin) URL (including [blob:](https://w3c.github.io/FileAPI/#DefinitionOfScheme) URLs) can be used. [data:](https://www.rfc-editor.org/rfc/rfc2397#section-2) URLs can also be used, but they create a worker with an [opaque origin](https://html.spec.whatwg.org/multipage/browsers.html#concept-origin-opaque).

As you can see  [on mdn](https://developer.mozilla.org/en-US/docs/Web/API/Worker#constructors), building a worker from a blob url is implemented in all browsers.

PR #978, opened 2023-03-01
ESM support

### LINKED ISSUE #977: ESM support
In a node.js environment, Papa parse only supports commonjs. I will create a PR to add ESM support.

btakita:
fixes https://github.com/mholt/PapaParse/issues/977

### DIFF
--- a/esm/index.browser.mjs
+++ b/esm/index.browser.mjs
@@ -1,0 +1,2 @@
+import '../papaparse.js';
+export default globalThis.Papa;
--- a/esm/index.mjs
+++ b/esm/index.mjs
@@ -1,0 +1,3 @@
+import { createRequire } from 'module';
+const require = createRequire(import.meta.url);
+export default require('../papaparse.js');
--- a/esm/package.json
+++ b/esm/package.json
@@ -1,0 +1,4 @@
+{
+	"main": "./index.mjs",
+	"browser": "./index.browser.mjs"
+}
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
 	},
 	"license": "MIT",
 	"main": "papaparse.js",
+	"module": "papaparse.mjs",
 	"browser": "papaparse.min.js",
 	"devDependencies": {
 		"chai": "^4.2.0",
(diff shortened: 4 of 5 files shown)


btakita:
This PR does not work in browsers. I will look into a solution to support browsers as well.

btakita:
With the latest force push, I added the `esm` directory, which supports esm + browser & esm + module exports.

btakita:
With the latest force push, I added the esm directory, which supports esm + browser & esm + module exports, using esm/package.json. ESM & vite browser builds are now supported.

```json
{
	"main": "./index.mjs",
	"browser": "./index.browser.mjs"
}
```

"browser" is more specific than "main" but less specific than "module", which is why I'm using "main" & "browser".

btakita:
Here is my astro test page:

```astro
---
---
<script>
	import Papa from 'papaparse'
	console.debug('/|debug|1', { Papa, 'globalThis.Papa': globalThis.Papa })
</script>
```

Which has the console output:

image

btakita:
There's a problem with the latest force pushed commit. The vite server side build seems to use the "browser" import, which causes an error:

```
ReferenceError: require is not defined
    at new DuplexStreamStreamer (/@fs/home/brian/work/portfoliome/censible-core/vendor/PapaParse/papaparse.js:920:16)
```

Looking into another fix for this...

btakita:
If vite fixes the issue https://github.com/vitejs/vite/issues/12253, then this PR should work.

In the meantime, I added https://github.com/btakita/PapaParse/tree/vite-plugin-top-level-await, which will work in server side vite builds using the vite-plugin-top-level-await plugin package.

JonET:
Would this allow us to use esm modules in angular-cli as well? Or would more work need to be done for that? Trying to avoid the "optimization bailouts" warning.

jschang19 (approved):
This modification works for my Nuxt3 project deploying on Cloudflare worker, where I couldn't deploy LLaMAindex since the current version doesn't come with an ESM support. Thx for your contribution!

montella1507:
So? Can we after years get ESM build?

SNThrailkill:
Is there more discussion to be had on this? It seems like a slam dunk and would greatly help a lot of developers. Can we please get this merged in?

pkurcx:
Maybe in 2026?

PR #986, opened 2023-03-23
Remove jsperf.com links from README.md.

simi:
- service is not available anymore

If the code hosted at those links is available anywhere, I can try to help to port it to other service (like http://jsben.ch/, https://perf.link/, https://jsbench.me/, ...).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,11 @@
 Parse CSV with JavaScript
 ========================================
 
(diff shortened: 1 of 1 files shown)


pokoli:
Thanks for your contribution. 

I do not think we have the code available somewhere else.

gabe565:
It looks like jsPerf has moved from jsperf.com to jsperf.app. I'm not sure if they still work as intended, but here are the new benchmark links:

- https://jsperf.app/javascript-csv-parsers/4
- https://jsperf.app/javascript-csv-parsers/3

PR #987, opened 2023-03-24
Fix for polluted Array

ponny:
We have some helper methods on our Array object.  Papaparse started blowing up with the latest update (seems to be from [this commit](https://github.com/mholt/PapaParse/commit/c1cbe16636be146f5f1d321cdd7cbddc9143b045)).

This bugfix updates the for loop to only interate over the array's elements, not the properties of the Array object.

Here's my testing sandbox, if you want to have a play: https://codesandbox.io/s/zealous-liskov-nuj41l?file=/src/index.js

I don't contribute to OSS often so this is quite new to me.  I'm happy to take any feedback on my pull.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -1476,7 +1476,7 @@ License: MIT
 				var headerCount = {};
 				var duplicateHeaders = false;
 
-				for (var j in headers) {
+				for (var j = 0; j < headers.length; j++) {
 					var header = headers[j];
 					if (isFunction(config.transformHeader))
 						header = config.transformHeader(header, j);
--- a/papaparse.js
+++ b/papaparse.js
@@ -1476,6 +1476,7 @@ License: MIT
 				var headerCount = {};
 				var duplicateHeaders = false;
 
+				// Using old-style 'for' loop to avoid prototype pollution that would be picked up with 'var j in headers'
 				for (var j = 0; j < headers.length; j++) {
 					var header = headers[j];
 					if (isFunction(config.transformHeader))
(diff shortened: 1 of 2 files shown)


pokoli:
I will be happy to include this, but just without the test suite. 
I do not think we want to explicity support the other properties but I can leave with the different for loop if that works for it. 
Could you add a comment in the code explaining why it is needed? Thanks.

ponny:
> I will be happy to include this, but just without the test suite. I do not think we want to explicity support the other properties but I can leave with the different for loop if that works for it. Could you add a comment in the code explaining why it is needed? Thanks.

Okay, it's your lib so I'll do it if you insist. Hear me out first though... :-)

Adding some methods to the Array prototype is a fairly common pattern.  It's how some old libaries like [Prototype](http://prototypejs.org/) work plus there's a a _heap_ of examples on StackOverflow where they suggest [polyfilling the Array prototype](https://stackoverflow.com/questions/51244103/confused-about-polyfill-of-array-prototype-reduce).   Also I worry that if I yank the test case, another loop could be added and break the functionality again.

renatodeleao:
Hello 👋! I think this chapter from the MDN docs explains the why very well:

> [Array iteration and for...in](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in#array_iteration_and_for...in)
> (...) However, the for...in loop will return all enumerable properties, **including those with non–integer names and those that are inherited**. Unlike for...of, for...in uses property enumeration instead of the array's iterator.

> **It is better to use a for loop with a numeric index**, Array.prototype.forEach(), or the for...of loop, because they will return the index as a number instead of a string, and also **avoid non-index properties**

Cheers!

ponny:
Righto, here you go.

Issue #985, opened 2023-03-23
Empty trailing fields populated with indexes after resuming a paused parser, loses place


landisdesign:
When using `parser.pause()`/`parser.resume()`, after resuming the parser, trailing empty cells in the row returned by `step` have `"_1"`, `"_2"`, `"_3"` etc. instead of empty cells. Thereafter, parsing appears to hit wrong locations.

Code sandbox here:
https://codesandbox.io/s/heuristic-violet-8wrinv?file=/src/CSVParser.js

pokoli:
It should be fixed with https://github.com/mholt/PapaParse/commit/824bbd9daf17168bddfc5485066771453cab423e
Please reopen if you do not agree.

landisdesign:
I replaced the package dependency with the new file from `main` in the CodeSandbox above, and it's returning the same results.

The odd thing is that, in addition to the underscore indices appearing, it looks like the parser is losing its place. Fields are getting eaten and the results are shifting left.

<img width="415" alt="Screenshot from code sandbox" src="https://user-images.githubusercontent.com/44303132/227247694-7bcb6cf8-cd90-413d-a01b-147277b78704.png">

pokoli:
Please provide a simple example that we can run on our test suite.

landisdesign:
I don't have a lot of Node experience (I'm a front-end developer 🤷‍♂️) but below is a new entry for the `CUSTOM_TESTS` suite, based on the test CSV from the code sandbox above. It appears to illustrate the problem accurately.

```js
	{
		description: "Pause and resume works with headers and empty trailing fields (Regression Test for Bug #985)",
		expected: [["Column 1", "Column 2", "Column 3", "Column 4", "Column 5"], [
			{ "Column 1": "R1C1", "Column 2": "", "Column 3": "R1C3", "Column 4": "", "Column 5": "" },
			{ "Column 1": "R2C1", "Column 2": "", "Column 3": "", "Column 4": "", "Column 5": "" },
			{ "Column 1": "R3C1", "Column 2": "", "Column 3": "", "Column 4": "R3C4", "Column 5": "" },
			{ "Column 1": "R4C1", "Column 2": "", "Column 3": "", "Column 4": "", "Column 5": "" },
		]],
		run: function(callback) {
			var inputString = [
				"Column 1,Column 2,Column 3,Column 4,Column 5",
				"R1C1,,R1C3,,",
				"R2C1,,,,",
				"R3C1,,,R3C4,",
[...]

PR #989, opened 2023-03-24
Generate headers in `Parser`, based upon presence of `_fields`, without changing `_input` (fixes issue #985)

landisdesign:
Previously, headers were created in `ParseHandle`'s `processResults()`, but also managed by `Parser.parse()` when `baseIndex` was falsy.

When headers were requested and `baseIndex` was false, `parse()` would modify `_input` without adjusting `cursor`, causing the parser to mistake where it needed to continue parsing. This would cause further rows to be shifted. In addition, `baseIndex` could be `undefined` or `0`, depending on how it was called, causing this `cursor` shift to potentially happen twice.

This PR:

* Uses `_fields.length` as the source of truth for whether or not headers should be generated when `config.header` is truthy
* Consolidates header generation within `Parser` by passing `_fields` to the `Parser` constructor where the above test is performed
* Uses the same parsing algorithms used for row generation for headers
* Adds `fields` to results generated by `Parser.parse()` based upon `fields` being populated or not
* Updates `_fields` based upon the presence of `fields` in `Parser.parse()`'s results

Additional impacts:

* Because header generation is now performed within `pushRow()`, the header row is no longer returned as data by `Parser.parse()`. Instead, it is returned in `meta.fields` by `returnable()`. This also means we don't need to adjust `config.preview` to account for the header row.

### DIFF
--- a/tests/test-cases.js
+++ b/tests/test-cases.js
@@ -591,8 +591,11 @@ var CORE_PARSER_TESTS = [
 		input: 'A,A,A,A\n1,2,3,4',
 		config: { header: true },
 		expected: {
-			data: [['A', 'A_1', 'A_2', 'A_3'], ['1', '2', '3', '4']],
-			errors: []
+			data: [['1', '2', '3', '4']],
+			errors: [],
+			meta: {
+				fields: ['A', 'A_1', 'A_2', 'A_3']
+			}
 		}
 	},
 	{
@@ -600,8 +603,11 @@ var CORE_PARSER_TESTS = [
 		input: 'A,A,A,A\n1,2,3,4',
 		config: { header: true, transformHeader: function(header) { return header.toLowerCase(); } },
 		expected: {
-			data: [['a', 'a_1', 'a_2', 'a_3'], ['1', '2', '3', '4']],
-			errors: []
+			data: [['1', '2', '3', '4']],
+			errors: [],
+			meta: {
+				fields: ['a', 'a_1', 'a_2', 'a_3']
+			}
 		}
 	},
 	{
@@ -609,8 +615,11 @@ var CORE_PARSER_TESTS = [
 		input: 'c,c,c,c_1\n1,2,3,4',
 		config: { header: true },
 		expected: {
-			data: [['c', 'c_1', 'c_2', 'c_1_0'], ['1', '2', '3', '4']],
-			errors: []
+			data: [['1', '2', '3', '4']],
+			errors: [],
+			meta: {
+				fields: ['c', 'c_1', 'c_2', 'c_1_0']
+			}
 		}
 	},
 ];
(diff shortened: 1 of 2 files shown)


tests/test-cases.js
landisdesign, line 1: Changes to existing tests were required to reflect that:
* `Parser` now requires an array as its second argument for its constructor
* `Parser.parse()` no longer returns the header row as data

beckyconning:
🎉

beckyconning:
Hey @landisdesign

Can't thank you enough.

PapaParse basically didn't work from in a stream pipeline without this.

You are a life saver.

mholt:
As long as it works, and the tests pass, this is OK with me -- will probably defer to @pokoli for a closer look though. Ping me again if it isn't merged in a while.

tests/test-cases.js
```diff
@@ -2018,6 +2027,51 @@ describe('Unparse Tests', function() {
 
 
 var CUSTOM_TESTS = [
+	{
```
pokoli, line 2030: Could you please add new tests at the end?

tests/test-cases.js
```diff
@@ -2018,6 +2027,51 @@ describe('Unparse Tests', function() {
...
+		description: "Pause and resume works with headers and duplicate fields (Regression Test for Bug #985)",
+		expected: [[
+			["Column 1", "Column 2", "Column 3", "Column 4", "Column 5"],
```
pokoli, line 2033: Do we really need to have 5 rows? Can we make the test simpler with less rows?

tests/test-cases.js
```diff
@@ -2018,6 +2027,51 @@ describe('Unparse Tests', function() {
...
+			].join("\n");
+			var output = [];
+			var dataRows = [];
+			var headerResults = [];
+			Papa.parse(inputString, {
+				header: true,
+				step: function(results, parser) {
+					if (results)
+					{
+						headerResults.push(results.meta.fields);
```
pokoli, line 2059: Why pusshing headers multiple times? Won't be better to just push when empty?

pokoli (changes requested):
The changes also look good to me, I added some comments to simplify the the tests a little bit. Could you have a look at it?

tests/test-cases.js
landisdesign: Yup! Done.

tests/test-cases.js
landisdesign: Probably not. I reduced it to two rows and four columns, enough to make sure we could test column value duplicates and that the headers didn't shift between pause/resume stages.

tests/test-cases.js
landisdesign: I wanted to double-check that multiple passes returned the same headers. It looked like the original code could have updated the headers on each pass, which seemed problematic. Perhaps that question makes sense as a separate test, or do you think it isn't worth the check?

PR #996, opened 2023-04-08
Update lovers.js

janswist:
Please review my addition of DocTemple service that is using PapaParse since the beginning as its go-to CSV parsing solution.

### DIFF
--- a/docs/resources/js/lovers.js
+++ b/docs/resources/js/lovers.js
@@ -27,6 +27,12 @@
 
 var peopleLovePapa = [
 	{
+		link: "http://doctempleapp.com",
+		name: "DocTemple",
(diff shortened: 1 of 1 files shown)


pokoli:
Hi @janswist ,

I see the core of your product is to import data using our library and that you are having a profit from it.
I'm wondering if you have any plans to support the maintenance of Paparse (either by money or by code). If yes, could you please share them?

janswist:
Hi @pokoli,

I'll be glad to contribute money whenever I will get first recurring customer. Even though it's already 2 years old it's quite hard to market such a specialized product and, to be honest, didn't had much success so far.

Lately I did a light design refresh and now looking for places where I can find potential users that actually need it.

So yes, I'm happy to share the success when there is one.

PR #997, opened 2023-04-12
Fixed cursor when duplicated header

jsakaicribl:
When processing duplicate headers with header auto, the position of meta.cursor will be returned as the position of cursor including renamed headers such as _<number>. But it's different from the cursor of the actual file. So set the cursor to the length of the original header.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -1469,9 +1469,10 @@ License: MIT
 				return returnable();
 
 			// Rename headers if there are duplicates
+			var firstLine;
 			if (config.header && !baseIndex)
 			{
-				var firstLine = input.split(newline)[0];
+				firstLine = input.split(newline)[0];
 				var headers = firstLine.split(delim);
 				var separator = '_';
 				var headerMap = new Set();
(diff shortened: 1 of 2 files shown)


papaparse.js
```diff
@@ -1516,7 +1517,11 @@ License: MIT
 				for (var i = 0; i < rows.length; i++)
 				{
 					row = rows[i];
-					cursor += row.length;
+					if (config.header && i === 0 && firstLine !== undefined) {
```
pokoli, line 1520: Do we really need to test if confix header? firstline will be undefined if config.header is not set, so I think we can remove the `config.header` test

pokoli, line 1520: Can you add a comment like: 
```
// use firstline as row length may be changed due to duplicated headers
```
before the if clause

tests/test-cases.js
```diff
@@ -2716,7 +2725,7 @@ describe('Custom Tests', function() {
 		const blob = new Blob([
 			`
 				importScripts('${papaParseScriptPath}');
-				
+
```
pokoli, line 2728: This seems unreleated to the change. Could you please avoid modifyinng unrelated lines? Thanks

pokoli (changes requested):

papaparse.js
jsakaicribl: Thanks. I fixed that.

jsakaicribl: Thanks. I fixed that.

pokoli:
Your patch is merged, thanks for your contribution!

PR #1017, opened 2023-08-22
Better handling of duplicate headers when parsing chunks

stuart-marshall:
The previous [fix](https://github.com/mholt/PapaParse/pull/997) for cursor management when there are duplicate headers worked for the faseMode parse path, but not for the slow mode parse path (that handles quotes).

This change is more general and works for both parse paths.

I also spotted a bug in handling quotes at the end of the file when the headers change.

Added new tests for both of these issues.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -1463,7 +1463,7 @@ License: MIT
 
 			// Establish starting state
 			cursor = 0;
-			var data = [], errors = [], row = [], lastCursor = 0;
+			var data = [], errors = [], row = [], lastCursor = 0, inputExpansion = 0;
 
 			if (!input)
 				return returnable();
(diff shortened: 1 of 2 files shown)


tests/node-tests.js
```diff
@@ -164,6 +164,46 @@ describe('PapaParse', function() {
...
+				{
+					if (stepped > 1) {
+						const startsWithEtiam = results.data && results.data.placeholder && results.data.placeholder.startsWith("Etiam");
+						const startsWithLorem = results.data && results.data.placeholder && results.data.placeholder.startsWith("Lorem");
+						startsWithEtiamOrLorem = startsWithEtiamOrLorem && (startsWithEtiam || startsWithLorem);
+					}
+				}
+			},
+			complete: function() {
+				assert(startsWithEtiamOrLorem);
```
pokoli, line 188: I'm not sure which is the objective of such test. 
Could add more assertions? Can we just test the stepped variable to ensure it has steeped correctly?

tests/node-tests.js
```diff
@@ -164,6 +164,46 @@ describe('PapaParse', function() {
 		});
 	});
 
+	it('Checks cursor when file is large and has duplicate headers', function(done) {
+		this.timeout(30000);
```
pokoli, line 168: Why is the timeout required? I think it should be removed.

papaparse.js
```diff
@@ -1508,7 +1508,17 @@ License: MIT
 				if (duplicateHeaders) {
 					var editedInput = input.split(newline);
 					editedInput[0] = Array.from(headerMap).join(delim);
+					// If we change the size of the input due to duplicate headers
```
pokoli, line 1511: Could we make the comment smaller? 

What about: `store the number of removed chars to correctly report meta.cursor` ?

papaparse.js
```diff
@@ -1517,12 +1527,7 @@ License: MIT
 				for (var i = 0; i < rows.length; i++)
 				{
 					row = rows[i];
-					// use firstline as row length may be changed due to duplicated headers
-					if (i === 0 && firstLine !== undefined) {
```
pokoli, line 1521: As firstline is no longer used, probably it can be removed as variable

pokoli (changes requested):
Thanks for your PR. I've added some comments. 
It is not clear for me what the new test is doing. IT will be great if you can explain it.

PR #1020, opened 2023-09-18
Update lovers.js

devinschumacher:


pokoli:
Please include your tool in the list intead of replacing your example. 

Also please provide a brief description why you are using Papaparse

Issue #738, opened 2019-11-14
Skip lines future (Feature Suggestion)


DamirKh:
It would be great to have the option "skip N lines", since CSV files often contain not commented headers. Can this be implemented?
For example:
==============================
This is a data file generated by some old software.
Next line will contain a headers of parameters.
Temperature, Humidity, Voltage
22.5, 45.5, 220
23.0, 44.0, 219
...
In such case "skipLines: 2" could be added to config.

Thanks!

pokoli:
Yes, it makes sense to me. It will be great if you provide a PR adding support for it. Please also include a test case to ensure the behaviour is working correctly.

DamirKh:
My low English level does not allow me to be sure that I understood you correctly. I'm not a public person and my PR is limited to my friends only. What did You mean "include a test case"? Fork PapaParse and add a automatic test case? I can try, but I'm just started to learn JS.

pokoli:
Yes, you should include this kind of tests on test tsts-test-cases.js

I will only susgest to add the two of the three sugestesd tests as the first are the second are teting the same but only with a diferent behaviour. It only change the variable used.

janisdd:
Why not just parse all and then remove the first `N` lines of the result??

Here is a one line solution

```js
var result = Papa.parse('the csv data')
var N = 2

[...Array(N)].forEach(_ => result.data.pop())

//result.data has the first N=2 rows removed
```

websharik:
@jscheid, 1) To skip "header" (N lines).  2) To correct parsing with attribute "header". 3) For big files with streaming.

Now i have large file with 4 first rows is "header" (1 row column names, 3 rows props about columns), other data.

pokoli:
@Lonli-Lokli No, I won't merge the PR unless it is green and updated to latest version. 
So it needs some work in order to merge it.

pokoli:
Implemented in #1021 1021

video-recording-fabian:
@janisdd https://reza-reviewer.github.io/shattering/

PR #1021, opened 2023-09-24
Skip lines future (Feature Suggestion) #738

bhuvaneshwararaja:
### Description:
In this feature enhancement, Added the capability to skip a defined number of lines when converting a CSV (Comma-Separated Values) file to JSON format. This functionality offers users more flexibility when dealing with CSV files with header information or metadata at the beginning, allowing for a more efficient and precise conversion process.

### Changes Made:

1. Introduced a new configuration parameter, skipFirstNLines, which allows users to specify the number of lines to skip at the beginning of the CSV file during the conversion to JSON. This parameter supports values from 0 and above.
2. Implemented validation for the skipFirstNLines configuration parameter to ensure that it does not negatively impact the parsing process, even if it is set to 0 or given a negative value.
3. Enhanced the user interface of the player.html file by adding an input field that allows users to input the desired value for skipFirstNLines.
4.  Updated the player.js script to retrieve and utilize the skipFirstNLines input from the user and incorporate it into the parser's configuration.

### Test cases:
 
Included test cases to validate the functionality of the skipFirstNLines feature, ensuring that it behaves correctly under various scenarios.

Test 1:  "Skip First N number of lines , with header",

Test 2: "Skip First N number of lines , without header",

Test 3: "Skip  First N lines with value as 0 with header false",

Test 4:  "Skip First N lines with value as 0 with header true",

Test 5: "Skip First N lines with value less then 0, with header true",
		
Test 6: "Without Skip First N Lines config",

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -484,6 +484,7 @@ License: MIT
 		}
 	}
 
+
 	/** ChunkStreamer is the base prototype for various streamer implementations. */
 	function ChunkStreamer(config)
 	{
@@ -519,8 +520,23 @@ License: MIT
 
 			// Rejoin the line we likely just split in two by chunking the file
 			var aggregate = this._partialLine + chunk;
+			this._pendingSkip = parseInt(this._config.skipFirstNLines) || 0;
+			this._skipHeader = 0;
+			if (this._config.header) {
+				this._skipHeader++;
+			}
+			if (this._pendingSkip > 0) {
+				var splitChunk = aggregate.split('\n');
+				var currentChunkLength = splitChunk.length;
+				if (currentChunkLength <= this._pendingSkip) {
+					aggregate = this._partialLine;
+				}
+				else{
+					aggregate = this._partialLine + [...splitChunk.slice(0, this._skipHeader), ...splitChunk.slice(this._skipHeader + this._pendingSkip)].join('\n');
+				}
+				this._pendingSkip -= currentChunkLength;
+			}
 			this._partialLine = '';
-
 			var results = this._handle.parse(aggregate, this._baseIndex, !this._finished);
 
 			if (this._handle.paused() || this._handle.aborted()) {
@@ -1929,7 +1945,6 @@ License: MIT
 	{
 		return function() { f.apply(self, arguments); };
 	}
-
 	function isFunction(func)
 	{
 		return typeof func === 'function';
--- a/tests/test-cases.js
+++ b/tests/test-cases.js
@@ -1574,6 +1574,42 @@ var PARSE_TESTS = [
 			data: [['a', 'b', 'c\n'], ['d', 'e', 'f']],
 			errors: []
 		}
+	},
+	{
+		description: "Skip First N number of lines , with header and 2 rows",
+		input: 'a,b,c,d\n1,2,3,4',
+		config: { header: true, skipFirstNLines: 1 },
+		expected: {
+			data: [],
+			errors: []
+		}
+	},
+	{
+		description: "Skip First N number of lines , with header and 3 rows",
+		input: 'a,b,c,d\n1,2,3,4\n4,5,6,7',
+		config: { header: true, skipFirstNLines: 1 },
+		expected: {
+			data: [{a: '4', b: '5', c: '6', d: '7'}],
+			errors: []
+		}
+	},
+	{
+		description: "Skip First N number of lines , with header false",
+		input: 'a,b,c,d\n1,2,3,4\n4,5,6,7',
+		config: { header: false, skipFirstNLines: 1 },
+		expected: {
+			data: [['1','2','3','4'],['4','5','6','7']],
+			errors: []
+		}
+	},
+	{
+		description: "Skip First N number of lines , with header false and skipFirstNLines as negative value",
+		input: 'a,b,c,d\n1,2,3,4\n4,5,6,7',
+		config: { header: false, skipFirstNLines: -2 },
+		expected: {
+			data: [['a','b','c','d'],['1','2','3','4'],['4','5','6','7']],
+			errors: []
+		}
 	}
 ];
 
--- a/docs/docs.html
+++ b/docs/docs.html
@@ -450,7 +450,8 @@ var csv = Papa.unparse({
 	beforeFirstChunk: undefined,
 	withCredentials: undefined,
 	transform: undefined,
-	delimitersToGuess: [',', '\t', '|', ';', <a href="#readonly">Papa.RECORD_SEP</a>, <a href="#readonly">Papa.UNIT_SEP</a>]
+	delimitersToGuess: [',', '\t', '|', ';', <a href="#readonly">Papa.RECORD_SEP</a>, <a href="#readonly">Papa.UNIT_SEP</a>],
+	skipFirstNLines: 0
 }</code></pre>
 					</div>
 					<div class="clear"></div>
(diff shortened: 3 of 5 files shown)


pokoli:
What you are proposing is already implemented using the `preview` config option.

bhuvaneshwararaja:
Hi @pokoli , 
I have gone through the documentation (https://www.papaparse.com/docs) for the  preview options 
and got to know that will help to parse the expected number of  rows in csv
example: If we have 20 rows in csv, If the preview value is 10 . Out of 20 only 10 rows will be parsed . 

But this PR will add a config to skip the N number of rows in the beginning of the csv.
Example: If we have 20 rows in csv , If skipNLines value is 10 . Out of 20 , first 10 rows will be skipped .

Please check it once again , it is relevant to the issue #738 as well.

pokoli:
Ok, then we need to document this new flag to explain which is the ussage. 

I think the flag will be better named as `skipFirstLines`. 

Could you have a look at the test suite? it seems you introduced some failing tests.

bhuvaneshwararaja:
Hi @pokoli , Thanks 
Sure i will check those test suite , I will update the PR with doc

bhuvaneshwararaja:
Hi @pokoli , I have updated the PR

papaparse.js
```diff
@@ -484,6 +484,10 @@ License: MIT
 		}
 	}
 
+	function customSplice(inputArray, startIndex, count) {
```
pokoli, line 487: Why is required? It will be better to just use the slice directly instead of a custom function

tests/test-cases.js
```diff
@@ -1574,6 +1574,60 @@ var PARSE_TESTS = [
...
+		description: "Skip First N number of lines , with header false and skipFirstNLines as negative value",
+		input: 'a,b,c,d\n1,2,3,4\n4,5,6,7',
+		config: { header: false, skipFirstNLines: -2 },
+		expected: {
+			data: [['a','b','c','d'],['1','2','3','4'],['4','5','6','7']],
+			errors: []
+		}
+	},
+	{
+		description: "Without Skip First N number of lines",
```
pokoli, line 1624: This test is not required as it is already tested by others.

tests/test-cases.js
```diff
@@ -1574,6 +1574,60 @@ var PARSE_TESTS = [
...
+		description: "Skip First N number of lines , with header false",
+		input: 'a,b,c,d\n1,2,3,4\n4,5,6,7',
+		config: { header: false, skipFirstNLines: 1 },
+		expected: {
+			data: [['1','2','3','4'],['4','5','6','7']],
+			errors: []
+		}
+	},
+	{
+		description: "Skip First N number of lines , with header false and skipFirstNLines as 0",
```
pokoli, line 1606: I do not think this test is required as this is the same as the default value.

papaparse.js
```diff
@@ -519,8 +523,21 @@ License: MIT
 
 			// Rejoin the line we likely just split in two by chunking the file
 			var aggregate = this._partialLine + chunk;
+			this._skipLines = parseInt(this._config.skipFirstNLines) || 0;
+			this._pendingSkip = parseInt(this._config.skipFirstNLines) || 0;
+			this._skipHeader = this._config.header ? 1 : 0;
```
pokoli, line 528: Wont' be simpler to just increse lines to skip by one when we have a header?

papaparse.js
```diff
@@ -519,8 +523,21 @@ License: MIT
 
 			// Rejoin the line we likely just split in two by chunking the file
 			var aggregate = this._partialLine + chunk;
+			this._skipLines = parseInt(this._config.skipFirstNLines) || 0;
+			this._pendingSkip = parseInt(this._config.skipFirstNLines) || 0;
```
pokoli, line 527: I think we do not need to have  the `_skipLines` variable but just the pendingSkip and deacrese it each time. 
When pendingSkip is lower or equal than zero we should not skip anything else. 

This will make the code simpler.

pokoli (changes requested):

tests/test-cases.js
bhuvaneshwararaja: This has been removed

papaparse.js
bhuvaneshwararaja: This has been fixed

pokoli (approved):

janisdd:
Just a small note: This does not work for quoted fields...

```
1,"B\nB",C,A,"B\nB",C,A,"B\nB",C,A,"B\nB",C
2,"B\nB",C,A,"B\nB",C,A,"B\nB",C,A,"B\nB",C
3,"B\nB",C,A,"B\nB",C,A,"B\nB",C,A,"B\nB",C
4,"B\nB",C,A,"B\nB",C,A,"B\nB",C,A,"B\nB",C
5,"B\nB",C,A,"B\nB",C,A,"B\nB",C,A,"B\nB",C
```
5 lines with 12 columns

```js
Papa.parse('1,"B\nB",C,A,"B\nB",C,A,"B\nB",C,A,"B\nB",C\n2,"B\nB",C,A,"B\nB",C,A,"B\nB",C,A,"B\nB",C\n3,"B\nB",C,A,"B\nB",C,A,"B\nB",C,A,"B\nB",C\n4,"B\nB",C,A,"B\nB",C,A,"B\nB",C,A,"B\nB",C\n5,"B\nB",C,A,"B\nB",C,A,"B\nB",C,A,"B\nB",C', 
{skipFirstNLines: 5})
```

skips the "real" first line (5 because \n after each line).

pokoli:
@janisdd It will be great if you can fill an issue/merge request to make this work with quotes also.

janisdd:
Well, no... I don't think this feature should be added to the library. Here are my thoughts:

- it is not clear how it interacts with other settings (e.g. header, skipEmptyLines, to be fair, the interaction of other options is not good documented either)
- can be solved by 1 line of code (see issue) and also works for quoted fields (via parsing everything)
- even if implemented to work with quoted fields, one has to parse the first N lines and then throw them away
  - this is because we only know it's a quoted field when fully parsed
  - so, skipping is essentially just parsing + discarding -> can be done by the user (in 1 line)

However, here are some notes if someone wants to implement it anyway:

- have a look at how the setting `preview` is implemented
- ... essentially just not add the `row` in https://github.com/mholt/PapaParse/blob/e2d570e77cc2720b2c256ec099f3e7e424c39e35/papaparse.js#L1742 as long as some skip counter is too small

I'm not sure if this would also work for streaming/chunking (or whatever, I just use the normal parsing). But as far as i know there is only one core logic/function for the actual parsing... which means this should also work for streaming

jkruke:
I think this feature hasn't been implemented as wanted in the referenced issue #738 !
The example CSV was:
```
This is a data file generated by some old software.
Next line will contain a headers of parameters.
Temperature, Humidity, Voltage
22.5, 45.5, 220
23.0, 44.0, 219
```
Expected output should be (to be fair, it wasn't precisely specified):
```
Temperature, Humidity, Voltage
22.5, 45.5, 220
23.0, 44.0, 219
``
But the actual output with `skipFirstNLines=2` is (according to the test cases):
```
This is a data file generated by some old software.
23.0
```
Analogue to the following test case in the code:
https://github.com/mholt/PapaParse/pull/1021/files#diff-e0ce8cb4901057c1880bee545909a64f38c7383b4d41982d6f2db9a8ec81eac7R1588
```{
		description: "Skip First N number of lines , with header and 3 rows",
		input: 'a,b,c,d\n1,2,3,4\n4,5,6,7',
		config: { header: true, skipFirstNLines: 1 },
		expected: {
			data: [{a: '4', b: '5', c: '6', d: '7'}],
			errors: []
		}
	}
```
This test case is not realistic because it does not reflect a CSV with some preamble lines to be skipped. 
should be rather: `input: 'to-be-ignored\na,b,c,d\n1,2,3,4',` and `expected.data: [{a: '1', b: '2', c: '3', d: '4'}]`.

If someone wanted to skip _rows_ of the record set, it should not be done during the parsing phase, but later when working with the data sets by doing some simple postprocessing such as `data = data.slice(1)`.

@bhuvaneshwararaja, @pokoli  could you please take a look at it and give me some feedback about my thoughts? :)

PR #1022, opened 2023-09-27
docs: fix default value of `header` to be `true`

eyalch:
The default value of `header` in the config object is `true`, but the docs claim that it's `false`.

Here's the source code of the actual default value:
https://github.com/mholt/PapaParse/blob/841e1d420dd2758cccb60a2bc1f2fcffd327c251/papaparse.js#L275

### DIFF
--- a/docs/docs.html
+++ b/docs/docs.html
@@ -430,7 +430,7 @@ var csv = Papa.unparse({
 	newline: "",	// auto-detect
 	quoteChar: '"',
 	escapeChar: '"',
-	header: false,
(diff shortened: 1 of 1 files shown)


mees-:
The code reference is to a function that converts JSON to csv. This doesn't mean that when converting from csv to JSON the default is `true`

dboskovic:
Closing this. The docs are correct in `unparse` https://github.com/mholt/PapaParse/blob/master/docs/docs.html#L260

PR #1026, opened 2023-10-25
strip BOM when streaming chunk

spearmootz:
when parsing a CSV and using step, we stream the file but papaparse does not handle the BOM like it does when it is handling a string.  so i moved the function stripBom to a place where it can be reused and then called the function on the first chunk only since it can only happen then.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -186,7 +186,13 @@ License: MIT
 	}
 
 
-
+	// Strip character from UTF-8 BOM encoded files that cause issue parsing the file
+	function stripBom(string) {
+		if (string.charCodeAt(0) === 0xfeff) {
+			return string.slice(1);
+		}
+		return string;
+	}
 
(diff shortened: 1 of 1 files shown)


papaparse.js
```diff
@@ -513,6 +511,7 @@ License: MIT
 			// First chunk pre-processing
 			if (this.isFirstChunk && isFunction(this._config.beforeFirstChunk))
 			{
+				chunk = stripBom(chunk);
```
pokoli, line 514: This only be called if befforeFisrtChunk is set, but we want to always strip bom. Could you fix it?

pokoli (changes requested):

papaparse.js
spearmootz: you are right. this is done

spearmootz: @pokoli can I get another review?

PR #1028, opened 2023-11-01
auto promise if no complete argument

vanodevium:
simple solution for automatic Promises

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -201,6 +201,10 @@ License: MIT
 
 		_config.transform = isFunction(_config.transform) ? _config.transform : false;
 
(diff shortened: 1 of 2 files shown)


dboskovic:
I wonder if there's any downside to just always returning a promise here?

vanodevium:
@dboskovic This is because the stream is returned otherwise.

dboskovic:
Ah right. In that case, can we add some errors if you happen to for example be using this in `step` mode which would be incompatible with the promise approach.

vanodevium:
@dboskovic 

I'm sorry, but after re-reading the documentation, I realized that promises cannot be made here.

PR #1041, opened 2024-02-19
Improve readability on dynamicTyping option

maon-fp:


pokoli:
Thanks for improving our docs!

PR #1044, opened 2024-03-04
Update papaparse link in package.json to use HTTPS

domdomegg:
This is used when showing the URL in NPM, and causes browser security warnings when clicking through to the site.


pokoli:
Merged it! Thank you so much!

Issue #1045, opened 2024-03-07
skipFirstNLines should not take the first line as header


jkruke:
I think this feature hasn't been implemented as wanted in the referenced issue #738 !
The example CSV was:

```
This is a data file generated by some old software.
Next line will contain a headers of parameters.
Temperature, Humidity, Voltage
22.5, 45.5, 220
23.0, 44.0, 219
```

Expected output should be (to be fair, it wasn't precisely specified):

```
Temperature, Humidity, Voltage
22.5, 45.5, 220
23.0, 44.0, 219
```

But the actual output with `skipFirstNLines=2` is (according to the test cases):

```
This is a data file generated by some old software.
23.0
```

Analogue to the following test case in the code:
https://github.com/mholt/PapaParse/pull/1021/files#diff-e0ce8cb4901057c1880bee545909a64f38c7383b4d41982d6f2db9a8ec81eac7R1588

```
{
		description: "Skip First N number of lines , with header and 3 rows",
		input: 'a,b,c,d\n1,2,3,4\n4,5,6,7',
		config: { header: true, skipFirstNLines: 1 },
		expected: {
			data: [{a: '4', b: '5', c: '6', d: '7'}],
			errors: []
		}
	}
```

This test case is not realistic because it does not reflect a CSV with some preamble lines to be skipped. 
should be rather: `input: 'to-be-ignored\na,b,c,d\n1,2,3,4',` and `expected.data: [{a: '1', b: '2', c: '3', d: '4'}]`.

If someone wanted to skip _rows_ of the record set, it should not be done during the parsing phase, but later when working with the data sets by doing some simple postprocessing such as `data = data.slice(1)`.

@bhuvaneshwararaja, @pokoli  could you please take a look at it and give me some feedback about my thoughts? :)

_Originally posted by @jkruke in https://github.com/mholt/PapaParse/issues/1021#issuecomment-1980130733_

PR #1046, opened 2024-03-07
Only skip first n lines in the first chunk and don't take the first line as header (#1045)

jkruke:
References #1045

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -513,8 +513,13 @@ License: MIT
 			// First chunk pre-processing
 			const skipFirstNLines = parseInt(this._config.skipFirstNLines) || 0;
 			if (this.isFirstChunk && skipFirstNLines > 0) {
-				const splitChunk = chunk.split('\n');
-				chunk = [...splitChunk.slice(skipFirstNLines)].join('\n');
+				let _newline = this._config.newline;
+				if (!_newline) {
+					const quoteChar = this._config.quoteChar || '"';
+					_newline = this._handle.guessLineEndings(chunk, quoteChar);
+				}
+				const splitChunk = chunk.split(_newline);
+				chunk = [...splitChunk.slice(skipFirstNLines)].join(_newline);
 			}
 			if (this.isFirstChunk && isFunction(this._config.beforeFirstChunk))
 			{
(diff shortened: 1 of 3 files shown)


tests/test-cases.js
```diff
@@ -1585,11 +1585,11 @@ var PARSE_TESTS = [
 		}
 	},
 	{
-		description: "Skip First N number of lines , with header and 3 rows",
```
pokoli, line 1588: Please do not remove existing test suite but add a new test latter.

papaparse.js
```diff
@@ -511,6 +511,11 @@ License: MIT
 		this.parseChunk = function(chunk, isFakeChunk)
 		{
 			// First chunk pre-processing
+			const skipFirstNLines = parseInt(this._config.skipFirstNLines) || 0;
+			if (this.isFirstChunk && skipFirstNLines > 0) {
+				const splitChunk = chunk.split('\n');
```
pokoli, line 516: Please do not hardcode the line break, there is a config setter which defines the valid line breaks and allows the user to change them. Code should work for any of them.

pokoli (changes requested):

papaparse.js
jkruke: Good point! I changed the code. Unfortunately, made it a bit more complicated because config.newline might be undefined. I used the same approach like [here](https://github.com/mholt/PapaParse/pull/1046/files#diff-fbb8d0b19dc4586fadd80f52a7b49415d2d41a5fdccaeaa728f7fca51cfbbb29R1090)

tests/test-cases.js
jkruke: I did'nt remove the test, I corrected it. 
As I already mentioned, the first line should be expected to be the header row in the result. (because it should be skipped with skipFirstNLines=1

foobarnes:
@jkruke @pokoli Is this ready to merge? Can I do any work on it to get it through? Pretty essentially on our end.

jkruke:
Actually, the reviewer @pokoli should re-review to my actions regarding to his [comments](https://github.com/mholt/PapaParse/pull/1046#pullrequestreview-1931586045)  
From my point of view, no further work needs to be done then

foobarnes:
@pokoli Any update here?

jkruke:
To get this done, maybe @foobarnes you wanna approve+merge this PR instead? :)

foobarnes (approved):

foobarnes:
My bad, didn't realize I had permissions!

newbeelearn:
can we merge this?

kikelkik:
I'd love to merge this!

jkruke:
I'd love to merge this!

jkruke:
@pokoli (or @mholt): this PR cannot be merged without your action here

Issue #1052, opened 2024-05-01
Quoted header values containing commas and comprised of the same string aren't able to be parsed.


alliefitter:
This is a pretty weird corner case, so let me know if y'all need more detail. Given a sheet with headers named `"Bar, Baz"` and `"Spam, Baz"`, after splitting the header row on `,`, Papa will treat ` Baz"` as duplicate header, and append `_1` to the second instance of it in `headerMap`. Then while seemingly attempting to remediate duplicates,  the second header value will become `"Spam, Baz"_1`, and seems to break parsing fields later on. The following scirpt...

```
import Papa from 'papaparse'

console.log(Papa.parse('Foo,"Bar, Baz","Spam, Baz",Some,Other,Headers\n1,2,3,4,5,6', { header: true }))
```

... will print...

```
{
  data: [],
  errors: [
    {
      type: 'Quotes',
      code: 'InvalidQuotes',
      message: 'Trailing quote on quoted field is malformed',
      row: 0,
      index: 16
    },
    {
      type: 'Quotes',
      code: 'MissingQuotes',
      message: 'Quoted field unterminated',
      row: 0,
      index: 16
    }
  ],
  meta: {
    delimiter: ',',
    linebreak: '\n',
    aborted: false,
    truncated: false,
    cursor: 57,
    fields: [
      'Foo',
      'Bar, Baz',
      'Spam, Baz"_1,Some,Other,Headers\n1,2,3,4,5,6'
    ]
  }
}
```

I was going to submit a PR, but the code is a bit difficult to follow. If this will take some time for y'all to get to, just comment here, and I can spend some time on a PR.

tony-cocco:
Just spent 30minutes when this describes my issue as well. Can reproduce on the demo site as well. Should be noted that `headers: true` must be set.

My test snippet:
```
"Enum 1 (A, B, Other)","Enum 2 (C, D, Other)"
A,Other
```

FilippoSalvarani21:
I think I have the same issue https://github.com/mholt/PapaParse/issues/1055

any way to fix?

tony-cocco:
> I think I have the same issue #1055
> 
> any way to fix?

We control the template we're parsing, so we just removed the duplicate strings in the headers. Alternatively, you could do `headers: false` but might change your logic for the rest of your processing.

FilippoSalvarani21:
> > I think I have the same issue #1055
> > any way to fix?
> 
[...]

Issue #1007, opened 2023-06-29
Incorrect header detection in CSV parsing 5.4.1


lucas-inc-query:
I have encountered an issue with version 5.4.1 when parsing CSV headers. It seems that headers are being incorrectly parsed and appended with "_x".

To reproduce the issue, go to https://www.papaparse.com/demo, check the "Header row" box and parse the following string:

`"foo,dog","foo,cat"`

Inspecting the `results > meta > fields` property shows us the following array:

`fields: [ "foo,dog", 'foo_1,cat"' ]`

The 2 error messages are:

`message: "Trailing quote on quoted field is malformed"`
`message: "Quoted field unterminated"`

It seems that commas within the headers leads to the parser believing there duplicate header fields? This problem does not occur if the `header: true` option is omitted from the config.

faulpeltz:
Unfortunately this seems to be the case with all quoted duplicate headers:
``"bla","duplicate","duplicate","foo"``

Results in:
``fields: [ "bla", "duplicate",  "duplicate\"_1,\"foo"]``

* It works correctly when the quotes are removed
* Reverting to 5.3.2 also works (but has has no duplicate header renaming)

basememara:
Related to #998

ahoisl:
Looks like this issue got fixed with #1058 👏  At least the issue we were experiencing with 5.4.1 😄

PR #1058, opened 2024-06-12
Refactor header renaming logic to adress #1052, #1007

augustmarowski:
I refactored the header renaming logic to adress #1052, #1007
and updated test to match the improved renaming logic.


viralpickaxe:
+1 on this we're seeing this in production!

augustmarowski:
@pokoli @mholt Could you please review this PR?

PR #1065, opened 2024-08-27
Update lovers.js to include Chartbrew

razvanilin:


pokoli:
Hi @razvanilin ,

I see you have a comercial Saas that uses our library. Did you every considered contributing to the project to help on its maintenance? For me this will show some love for the project.

PR #1076, opened 2024-11-21
Fix improper handling of msg.results in chunk callbacks (#964)

### LINKED ISSUE #964: When using chunks callback, complete callback parameters are undefined
# Scenario
I'm attempting to use the `chunk` callback to obtain status updates on the parse operation for large files (similar to https://github.com/mholt/PapaParse/issues/385).  I noticed that the `step` callback is called for each line which makes it inefficient for calculating progress, so the `chunk` callback seems to be the best fit for this scenario.

# Problem
When calling `parse` with the `chunk` callback defined, both the `results` and the `file` parameter passed to the `complete` callback are undefined.  The  `results` and the `file` parameter passed to the `chunk` callback are defined as expected.

When calling `parse` without a `chunk` callback defined, both the `results` and the `file` parameter passed to the `complete` callback are defined as expected.

# Expected Behavior

Whether or not the `chunk` callback is defined, the `complete` callback should fire exactly once with the result set.

# Relevant Versions:
*  "@types/papaparse": "^5.3.5",
* "papaparse": "^5.3.2",
* "typescript": "~4.6.4",
* "@angular/common": "~13.3.0",

This seems to behave identically on the following browsers:
* Firefox 107.0.1
* Edge 108.0.1462.54

This occurs both when importing the raw PapaParse module via `import { parse }  from "node_modules/papaparse/papaparse.js"` as well as when importing the typed version via `import { parse }  from "papaparse"`

# Possibly related issues:
* https://github.com/mholt/PapaParse/issues/616
* https://github.com/mholt/PapaParse/issues/92
* https://github.com/mholt/PapaParse/issues/690
* https://github.com/mholt/PapaParse/issues/955
* https://github.com/mholt/PapaParse/issues/385
* https://github.com/mholt/PapaParse/issues/171
* https://github.com/mholt/PapaParse/issues/616

# Sample Code:

```TypeScript
    public ParseContactCSV(f: File, hasHeaderRow: boolean): Observable<ContactImportStatusUpdate> {
      const statusUpdateDebounceMS = 250
      let lastStatusTimestamp: number;
      let lastPercent = 0;
      return new Observable<ContactImportStatusUpdate>((o)=>{
        o.next({
          operation: "Parse Import", 
          statusText: "Beginning parse",
          result: null
        })
        parse<File>(f, {
          worker: true, // The chunk/complete callback bug happens whether or not we're using a worker
          header: hasHeaderRow,
          chunkSize: f.size/100, // The chunk/complete callback bug happens whether or not chunk size is defined.
          chunk(results, parser) {
            const percentDone = Math.ceil((results.meta.cursor / f.size)*100)
            // limit our status emits to one-percent steps and to emit no more frequently than every statusUpdateDebounceMS
            if (percentDone > lastPercent && (Date.now() - lastStatusTimestamp) <  statusUpdateDebounceMS) { 
              o.next({
                operation: "Parse Import", 
                statusText: "Parsed " + percentDone.toLocaleString(undefined,{maximumFractionDigits:2}) +"%",
                result: null
              })
              lastPercent = percentDone
              lastStatusTimestamp = Date.now();
            }
          },
          complete(results, file) {
            o.next({
              operation: "Parse Import", 
              statusText: "Complete; parsed " + results.data.length + " contacts",
              result: results
            })
          },
        })
      });
    }
```

jongwoo328:
Fix #964.

The `delete msg.results;` statement was likely introduced for memory optimization.
However, this caused the results to not be correctly passed as a parameter to `complete` when a chunk callback was present.

To address this, the release point of `msg.results` has been adjusted to ensure that results are properly passed to `complete` in cases where a chunk callback is used.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -1866,24 +1866,25 @@ License: MIT
 					if (aborted)
 						break;
 				}
-				delete msg.results;	// free memory ASAP
 			}
 			else if (isFunction(worker.userChunk))
 			{
 				worker.userChunk(msg.results, handle, msg.file);
-				delete msg.results;
 			}
 		}
 
 		if (msg.finished && !aborted)
-			completeWorker(msg.workerId, msg.results);
+			completeWorker(msg.workerId, msg.results, msg.file);
+
+		delete msg.results;
 	}
 
-	function completeWorker(workerId, results) {
+	function completeWorker(workerId, results, file) {
 		var worker = workers[workerId];
-		if (isFunction(worker.userComplete))
-			worker.userComplete(results);
-		worker.terminate();
+		if (isFunction(worker.userComplete)) {
+			worker.userComplete(results, file);
+			worker.terminate();
+		}
 		delete workers[workerId];
 	}
 


papaparse.js
```diff
@@ -1866,24 +1866,25 @@ License: MIT
...
+
+		delete msg.results;
 	}
 
-	function completeWorker(workerId, results) {
+	function completeWorker(workerId, results, file) {
 		var worker = workers[workerId];
-		if (isFunction(worker.userComplete))
-			worker.userComplete(results);
-		worker.terminate();
```
pokoli, line 1886: Could you elaborate why wee need to terminate the worker only when a function is defined?

pokoli (changes requested):
Thanks for submiting the PR. 

Do you think it should be easy to have a test suite which tests that the chunk Callback gets proper results? It will be great to include such test to prevent any regresion on the future.

papaparse.js
jongwoo328: I think I may have unintentionally changed the original flow while modifying the code. I'll revert back to. Thank you.

jongwoo328:
Sure.

Should I add test to the node_test.js and CUSTOM_TEST in test-cases.js?

pokoli:
> Should I add test to the node_test.js and CUSTOM_TEST in test-cases.js?
`test-cases.js` sounds like the right path for such test. 

I will wait for your test before merging this.

jongwoo328:
I've addressed my mistakes and adding tests. 
Please let me know if there's anything else that needs to be adjusted. 
Thank you for your guidance!

pokoli:
> I've addressed my mistakes and adding tests. Please let me know if there's anything else that needs to be adjusted. Thank you for your guidance!

I've run the test and I see your new test is failing. Could you have a look and address it? TIA

jongwoo328:
Previous fixes didn't work on node environments 😅.
When using `ChunkStreamer` in a node environment, the `_completeResults` value being passed to `complete` was being set to `undefined` if the user provided the `chunk` option (in `parseChunk`).
I added code to store the results in `_completeResults` before setting `results` to `undefined` so that result can be properly passed to `complete` even if the `chunk` option is given.

Now it passes all tests!

jongwoo328:
Hi @pokoli
just checking in to see if there’s any feedback or updates on this PR. Let me know if I should adjust anything. Thanks!

jongwoo328:
Hi @pokoli 
Could you please take another look when you have a moment?

Issue #1054, opened 2024-05-22
Step meta.cursor not correctly outputting current byte correctly when using parser.pause() 5.3.2


eeeBs:
Currently running v5.3.2 due to the CSV parsing being broken in 5.4.1 #998 

Is anyone else able to reproduce this?

``` const { size } = fs.statSync(filePath);
       const stepLogic = (results: any, parser: any) => {
            parser.pause();
            setTimeout(() => {
                parser.resume();
            }, 50)
            const size = parser.streamer._config.size
            console.log('Current Position is: ' + results.meta.cursor + ', Filesize is ' + size)
        }
        Papa.parse(fileData, {
            header: true,
            size,
            chunkSize: 500,  // Set chunkSize per recommendation
            step: stepLogic,
        }); 
```

which gets me an output of
        
``` Current Position is: 1278, Filesize is 2562317
Current Position is: 1289, Filesize is 2562317
Current Position is: 1275, Filesize is 2562317
Current Position is: 1364, Filesize is 2562317
Current Position is: 1289, Filesize is 2562317
Current Position is: 1303, Filesize is 2562317
Current Position is: 1295, Filesize is 2562317
Current Position is: 1357, Filesize is 2562317
Current Position is: 1313, Filesize is 2562317
Current Position is: 1328, Filesize is 2562317
Current Position is: 1281, Filesize is 2562317
Current Position is: 1293, Filesize is 2562317
Current Position is: 1285, Filesize is 2562317
Current Position is: 1282, Filesize is 2562317
Current Position is: 1301, Filesize is 2562317
Current Position is: 1267, Filesize is 2562317
Current Position is: 1277, Filesize is 2562317
Current Position is: 1323, Filesize is 2562317
Current Position is: 1395, Filesize is 2562317
Current Position is: 1358, Filesize is 2562317
Current Position is: 1334, Filesize is 2562317
Current Position is: 1308, Filesize is 2562317
Current Position is: 1289, Filesize is 2562317
Current Position is: 1302, Filesize is 2562317
Current Position is: 1289, Filesize is 2562317
Current Position is: 1296, Filesize is 2562317
Current Position is: 1286, Filesize is 2562317
Current Position is: 1343, Filesize is 2562317
 ```

Pretty sure this should be working, but maybe I am misunderstanding something?

Issue #1029, opened 2023-11-08
transformHeader called twice


kiocosta:
I have this sample code
```
let counter = 0;
let csv = "First name,Last name\nJohn,Doe"
let json = Papa.parse(csv, {
    header: true,
    transformHeader: (header, a) => {
        counter++;
        console.log(a);
        return header;
    },
 });
console.log({ counter })
```
This is the result from running it:
![image](https://github.com/mholt/PapaParse/assets/59743948/5f114472-8716-4896-b59e-d3e06233beb6)
It's interesting that the second argument starts as a string and then becomes a number. I think this suggests transformHeader is being executed from different lines each time.

Logging "header" returns this:
"First name"
"Last name"
"First name"
"Last name"

It seems like transformHeader is executed twice for the first row. This is causing a bug in my application and I don't know why this behavior is happening. 

I'd really appreciate some help with it. Thanks in advance!!

kiocosta:
I guess the example I've posted is working as expected.
In my application a bug was happening, but that was because I had duplicate headers (some headers ended up becoming just "" after my transformation). This was causing unexpected behaviors. After fixing this, the problem ceased to occur.

kiocosta:
I guess the example I've posted is working as expected.
In my application a bug was happening, but that was because I had duplicate headers (some headers ended up becoming just "" after my transformation). This was causing unexpected behaviors. After fixing this, the problem ceased to occur.

samy-mssi:
> I guess the example I've posted is working as expected. In my application a bug was happening, but that was because I had duplicate headers (some headers ended up becoming just "" after my transformation). This was causing unexpected behaviors. After fixing this, the problem ceased to occur.

Hi, I am sorry to open this issue again, but I am facing the same issue after upgrading my whole React project, and you are the only one that got the same behavior.

I have an array "headers", and I push each header on it.
```
      transformHeader: (header, index: number) => {
        console.log('headers', headers, index);
        const headerValue = () => {
          return header
            ? header
            : t('grid.header.auto-generated-label', {
                columIndex: index + 1,
              });
        };
        const name = fileWithHeader ? headerValue() : `${index + 1}`;
        headers.push(name);
        headers = utils.getDuplicatesHeader(headers);
        return name;
      },
```
and here is what I got in my console : 
![image](https://github.com/mholt/PapaParse/assets/151582118/97ff93c9-4956-4cb4-b002-a48570981242)

I checked my csv and of course I have only 8 columns...

[...]

Issue #1006, opened 2023-06-21
transformHeader returns function instead of string: trim is not a function


bettysteger:
In version 5.3.2 this works:

```
transformHeader: function(header) { return header.trim(); }
```

In version 5.4.1 i got the following error because header is sometimes a function, sometimes a string, but the 1st param should always be a string?!

```
TypeError
header.trim is not a function. (In 'header.trim()', 'header.trim' is undefined)
```

kraighamady:
You can try declaring the function elsewhere?  I was able to make it work with an arrow function instead of a declaration.

example:
```
transformHeader: (header) => { return header.trim() },
```
Hopefully that works for you?

bettysteger:
@kraighamady this makes no difference

kraighamady:
@bettysteger Two thing I just noticed.  With a type error you are using typescript?  if so, you can do what I did which is explicitly type the header param in the function:

```
transformHeader: (header: string) => { return header.trim() },
```

The issue is that your typescript checker doesn't know that header is a string so cannot be sure that the trim() function is by default in the prototype.  

I am not sure why you can get away without typing in the latest version unless they added stricter types to the library to assist?

bettysteger:
@kraighamady unfortunately I am not using typescript for this project

kraighamady:
I'd have to see more of the code and the values in your csv.  You could try forcing the variable to string:

```
String(header).trim()

```

bettysteger:
@kraighamady but why do i need that now? in v5.3.2 it works, because it is always a string, and somehow in v5.4.1 it seems that the function is called twice: https://github.com/mholt/PapaParse/issues/1029

kraighamady:
Not sure.  I am having the duplicate call issue as well.  I was just offering a workaround in case you were blocked.  I am just a user of the package and not involved in the development of the tool.

bettysteger:
@kraighamady oh thanks! no i just keep v5.3.2 for now :D

Issue #998, opened 2023-04-19
CSV parsing is broken in 5.4.1


rpr-ableton:
Hi everyone,

First of all, big thanks for this great library that's been a big help in my work :raised_hands: 

I noticed when going from 5.3.2 to 5.4.1 that some CSV files can't be parsed successfully anymore. I'll attach below a small CSV, a nodejs script to reproduce the issue as well as the output.

`example.csv`:
```csv
colName_2,colName_3,another_col,entity,some_bool,period,date,value_eur
"foo, bar","foo, bar",foo_bar_baz_V1,bar baz,0,2022_02,02/01/22,1234.45678
"foo, bar","foo, bar",foo_bar_baz_V1,bar baz,0,2022_03,03/01/22,1234.45678
"foo, bar","foo, bar",foo_bar_baz_V2,bar baz,0,2022_11,11/01/22,-12345.67891
"foo, bar","foo, bar",foo_bar_baz_V2,bar baz,0,2022_12,12/01/22,-12345.67891
```

`script.js`:
```javascript
const papa = require('papaparse');
const fs = require('fs');

const readStream = fs.createReadStream(`${ __dirname }/example.csv`);
return papa.parse(readStream, {
    step: (res, parser) => {
        parser.pause();
        Object.keys(res.data).
            sort().
            forEach((key) => {
                console.log(res.data[key]);
            });
        return parser.resume();
    }, // step
    header: true,
    error: (err) => {
        return console.error(err);
    }, // error
    complete: () => {
        console.log('finished');
    }, // complete
}); // return papa.parse
```

Output with papaparse 5.4.1:
```
$ node script.js 
foo_bar_baz_V1
foo, bar
foo, bar
02/01/22
bar baz
2022_02
0
1234.45678
[ '-12345.67891' ]
foo, bar
foo, bar
foo_1, bar"_1,foo_bar_baz_V1,bar baz,0,2022_03,03/01/22,1234.45678
"foo, bar
2022_11
foo_bar_baz_V2
0
bar baz
11/01/22
foo, bar"_1,foo_bar_baz_V2,bar baz,0,2022_12,12/01/22,-12345.67891


 bar"
finished
```

Let me know if you need any additional information to better capture the issue.

Thanks a bunch.

pokoli:
Hi @rpr-ableton 

Could you please describe which is the issue?

rpr-ableton:
The symptoms are:
- some values from the CSV are parsed as arrays containing a single value
[...]

PR #1080, opened 2025-01-09
Add allowDuplicateHeaders option to configuration and update tests

tnrich:
The automatic duplicate header parsing introduced in a recent version, while helpful for new development, has broken existing legacy code in our codebase.

This PR adds a flag `allowDuplicateHeaders` which can be set to true to revert to the earlier functionality (as seen in v5.2.0 - that's the version we were on)

Let me know if this PR needs anything else and thanks for the consideration! Love papaparse :)

### DIFF
--- a/docs/docs.html
+++ b/docs/docs.html
@@ -450,6 +450,7 @@ var csv = Papa.unparse({
 	beforeFirstChunk: undefined,
 	withCredentials: undefined,
 	transform: undefined,
+	allowDuplicateHeaders: undefined,
 	delimitersToGuess: [',', '\t', '|', ';', <a href="#readonly">Papa.RECORD_SEP</a>, <a href="#readonly">Papa.UNIT_SEP</a>],
 	skipFirstNLines: 0
 }</code></pre>
--- a/papaparse.js
+++ b/papaparse.js
@@ -1741,7 +1741,7 @@ License: MIT
 			/** Returns an object with the results, errors, and meta. */
 			function returnable(stopped)
 			{
-				if (config.header && !baseIndex && data.length)
+				if (config.header && !config.allowDuplicateHeaders && !baseIndex && data.length)
 				{
 					var result = data[0];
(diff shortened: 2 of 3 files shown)


pokoli:
A list of renamedHeaders is returned as part of the result, so you can read the values and keep the previous behaviour. 

You just need to adapt your legacy code to read this new data. 

There is nothing to do from the PapaParse side.

tnrich:
@pokoli that isn't entirely true.. the other benefit to my legacy code from this option would be to prevent transformHeader from being run multiple times for the same header, see - https://github.com/mholt/PapaParse/issues/1029

Unfortunately the code added to rename the headers causes the tranformHeader function to be called a 2nd time which breaks code in our codebase that is depending on that function being run only once per header (which I don't think was an unreasonable assumption and allowed us to do custom logic in our CSV parser)

Why not add this simple option to optionally undo a BREAKING change that was pushed in a MINOR version? This doesn't seem like an unreasonable request and I believe would give an immediate fix to a whole host of tickets related to the changes made after v5.3.2 - 

https://github.com/mholt/PapaParse/issues/1054 https://github.com/mholt/PapaParse/issues/1054 https://github.com/mholt/PapaParse/issues/998 https://github.com/mholt/PapaParse/issues/1006 

I'm not suggesting that those shouldn't be fixed but it isn't a good practice to push a breaking change in a minor version without any recourse for end users.

tnrich:
Hope the above comment doesn't come off as harsh. Appreciate your stewardship of this important repo @pokoli! Overall I agree that the header de-dupe is a step in the right direction, unfortunately a lot of legacy code _will_ be broken by the two changes (automatic header dedupe and additional calling of transformHeader).

Targzp:
I am disappointed that the author did not respond to your follow-up questions. I also encountered the same problem.

danieljbrooks-GitHub:
And I as well

PR #1086, opened 2025-01-27
Only attempt to parse headers once

### LINKED ISSUE #1085: Data corruption in 5.5.1 when streaming data with step function
We recently tried upgrading from 5.4.1 to 5.5.1 and encountered this error

We started seeing `Duplicate headers found and renamed` in the logs and noticed that our data was corrupted.

It seems that instead of deduplicating just the header row, it deduplicates all the data in each row too.

You can reproduce with this file:
```

const csv = `a,b,c,c
d,d,e,e
d,f,f,g`;

Papa.parse(csv, {
    header: true,
    step: function(results) {
        console.log(results.data);
    },
    complete: function() {
        console.log("Parsing complete");
    }
});
```

In 5.4.1 it outputs the following, correctly deduplicating the extra `c` column
```
{ a: 'd', b: 'd', c: 'e', c_1: 'e' }
{ a: 'd', b: 'f', c: 'f', c_1: 'g' }
Parsing complete
```

In 5.5.1 it outputs the following, incorrectly deduplicating the `d`, `e`, and `f` in the data

```
Duplicate headers found and renamed.
Duplicate headers found and renamed.
{ a: 'd', b: 'd_1', c: 'e', c_1: 'e_1' }
Duplicate headers found and renamed.
{ a: 'd', b: 'f', c: 'f_1', c_1: 'g' }
Parsing complete
```

guseggert:
Hopefully this is the right way to fix this, I'm not very familiar with this codebase so please advise :) .

fixes #1085

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -1426,6 +1426,7 @@ License: MIT
 		var fastMode = config.fastMode;
 		var quoteChar;
 		var renamedHeaders = null;
+		var headerParsed = false;
 
 		if (config.quoteChar === undefined || config.quoteChar === null) {
 			quoteChar = '"';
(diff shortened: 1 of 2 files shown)


pokoli:
Thanks, i merged it!

PR #961, opened 2022-11-23
Handle parsing utf-8 bom encoded files

peteruithoven:
Strip a character that's added in utf-8 BOM encoded files, fixing parsing issues. 

Fixes: https://github.com/mholt/PapaParse/issues/840
Possibly also fixes: https://github.com/mholt/PapaParse/issues/917

Issue #372, opened 2017-03-15
Unicode BOM messes up first property name


LeaVerou:
First off, thanks for the great library! Its API is a work of art. 👍 

I ran into an issue when parsing CSV files fetched from the Github API with `headers: true`. The first column didn't have the property name I expected due to an invisible character which turned out to be the UTF-8 BOM (`U+FEFF`). I fixed it in my code via:

```js
response = response.replace(/^\ufeff/, "");
```

However, I thought you guys might want to fix it in the library at some point (and include other BOMs too, like the UTF-16 BOM etc). It's debatable whether this should be fixed in the library, and I could argue it both ways, but thought I should bring it to your attention and let you decide. :)

pokoli:
Hi @LeaVerou,

Thanks for opening the issue. 

Which is the current behaviour? Is Papa crashing? Could you provide some example csv (or some test scenario better) in order to reproduce the issue.

LeaVerou:
No crashes, no errors, just the BOM becomes part of the first property name.

pokoli:
I think we can not do anything more. If the input contains this data we should show this data and the only solution is to fix the data before (or after) parsing.

cweezy:
@LeaVerou I was running into this problem too, and I fixed the problem a different way. I am using javascript's [`FileReader`](https://developer.mozilla.org/en-US/docs/Web/API/FileReader) API and was using `readAsBinaryString` on a CSV file. Switching to `readAsText` for CSV files got rid of these BOMs.

Not sure if this would fix your issue, but figured it was worth sharing.

havenl:
@cweezy `readAsText` apparently does not remove the BOM in Internet Explorer (I was testing in IE10, not sure about other versions), so you'll still want to manually remove it _just in case_ your user uses IE.

@pokoli I would argue the library should remove the BOM as other CSV parsers do not make prepend BOM, an invisible character, to the first property key.

pokoli:
@havenl I'm open for discussion about remove the BOM but if the problem is that Internet Explorer 10 is not properly handling the text stream I don't think we should do anything. Internet Explorer 10 is not supported by Microsoft since December 13, 2016, so you users should upgrade it's browser.

kaligrafy:
It happens if we use Papa with nodejs also if the csv files contain BOM.
It is not that trivial to remove BOM when using a readstream. Would it be possible to remove the DOM when parsing the header automatically?

PaulMest:
[...]

PR #1090, opened 2025-03-14
Strip BOM character from column keys when parsing header, add BOM tests

baumschubser:
Addresses #372, stripping BOM from column keys when parsing the header.
The problem: The first column key includes the invisible BOM character if it exists in the CSV.

Pull request #961 handled BOM stripping for parsing the CSV without keyed data. However, with `header: true` the issue still persisted. This is solved by this pull request that runs the `stripBom` function from #961 also on the header field, before 'transformHeader'.

Also, I added tests for BOM handling.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -185,8 +185,14 @@ License: MIT
 		global.onmessage = workerThreadReceivedMessage;
 	}
 
-
-
+	// Strip character from UTF-8 BOM encoded files that cause issue parsing the file
+	function stripBom(string) {
+		console.log(`checking ${string}`);
+		if (string.charCodeAt(0) === 0xfeff) {
+			return string.slice(1);
+		}
+		return string;
+	}
 
 	function CsvToJson(_input, _config)
 	{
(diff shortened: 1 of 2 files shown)


papaparse.js
```diff
@@ -185,8 +185,14 @@ License: MIT
 		global.onmessage = workerThreadReceivedMessage;
 	}
 
-
-
+	// Strip character from UTF-8 BOM encoded files that cause issue parsing the file
+	function stripBom(string) {
+		console.log(`checking ${string}`);
```
pokoli, line 190: Could you please remove console.logs?

tests/test-cases.js
```diff
@@ -1691,6 +1709,18 @@ var PARSE_ASYNC_TESTS = [
 			errors: []
 		}
 	},
+	{
+		description: "Simple file with BOM encoding and header",
```
pokoli, line 1713: I do not think this test is necessary as its already tested by previous ones.
So I think we can just remove it.

pokoli (changes requested):
Hi, 

Thanks for your proposal. There are some minors fixes that should be addressed before merging your changes. Would you mind fixing them by amending the changes to current commit?

baumschubser:
Oops, sorry for that. I removed the obsolete test and the console.log

pokoli:
I merged the changes, thanks for contributing in the project!

PR #1094, opened 2025-05-03
optimize CSV parsing for better performance

Ray0907:
- Pre-compile quote regex for better performance
- Replace substring comparisons with startsWith() for delimiter and newline checks
- Refactor extraSpaces() function to avoid string creation and use direct character code checks.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -1516,6 +1516,7 @@ License: MIT
 
 			var nextDelim = input.indexOf(delim, cursor);
 			var nextNewline = input.indexOf(newline, cursor);
+			// Pre-compile the regex for escaped quotes for performance
 			var quoteCharRegex = new RegExp(escapeRegExp(escapeChar) + escapeRegExp(quoteChar), 'g');
 			var quoteSearch = input.indexOf(quoteChar, cursor);
 
@@ -1584,7 +1585,8 @@ License: MIT
 						var spacesBetweenQuoteAndDelimiter = extraSpaces(checkUpTo);
 
 						// Closing quote followed by delimiter or 'unnecessary spaces + delimiter'
-						if (input.substr(quoteSearch + 1 + spacesBetweenQuoteAndDelimiter, delimLen) === delim)
+						// Use startsWith for potentially better performance and readability
+						if (input.startsWith(delim, quoteSearch + 1 + spacesBetweenQuoteAndDelimiter))
 						{
 							row.push(input.substring(cursor, quoteSearch).replace(quoteCharRegex, quoteChar));
 							cursor = quoteSearch + 1 + spacesBetweenQuoteAndDelimiter + delimLen;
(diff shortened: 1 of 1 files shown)


pokoli:
Thanks for the proposal. 

Could you please provide some benchmark that proves your proposal makes the code faster?

papaparse.js
```diff
@@ -1516,6 +1516,7 @@ License: MIT
 
 			var nextDelim = input.indexOf(delim, cursor);
 			var nextNewline = input.indexOf(newline, cursor);
+			// Pre-compile the regex for escaped quotes for performance
```
pokoli, line 1519: I do not see any difference despite the comment. So I understand nothing is changed here. Rirgh? Or what is your change?

papaparse.js
```diff
@@ -1584,7 +1585,8 @@ License: MIT
 						var spacesBetweenQuoteAndDelimiter = extraSpaces(checkUpTo);
 
 						// Closing quote followed by delimiter or 'unnecessary spaces + delimiter'
-						if (input.substr(quoteSearch + 1 + spacesBetweenQuoteAndDelimiter, delimLen) === delim)
+						// Use startsWith for potentially better performance and readability
```
pokoli, line 1588: There is no need to include such comment

papaparse.js
```diff
@@ -1693,17 +1696,37 @@ License: MIT
...
 			/**
-             * checks if there are extra spaces after closing quote and given index without any text
-             * if Yes, returns the number of spaces
-             */
+			 * Checks if there are insignificant whitespace characters after a closing quote and before
+			 * the next delimiter or newline. If so, returns the number of such characters.
+			 * This function has been optimized to avoid creating substrings for performance.
+			 */
 			function extraSpaces(index) {
+				// If index is -1, delimiter/newline wasn't found after the quote.
```
pokoli, line 1704: No need for the comment.

papaparse.js
```diff
@@ -1693,17 +1696,37 @@ License: MIT
...
+
 				var spaceLength = 0;
-				if (index !== -1) {
-					var textBetweenClosingQuoteAndIndex = input.substring(quoteSearch + 1, index);
-					if (textBetweenClosingQuoteAndIndex && textBetweenClosingQuoteAndIndex.trim() === '') {
-						spaceLength = textBetweenClosingQuoteAndIndex.length;
+				for (var i = start; i < index; i++) {
+					var charCode = input.charCodeAt(i);
+					// Check for space (32) or tab (9)
+					// This aims to match the common behavior implicitly handled by the previous .trim() check
```
pokoli, line 1720: I do not think we should reimplement trim function but just call it. Please keep the previous trimg function. 

There reason is that trim may remove more empty chars that I do not want to maintain as part of the library.

papaparse.js
```diff
@@ -1693,17 +1696,37 @@ License: MIT
...
+					if (charCode === 32 || charCode === 9) {
+						spaceLength++;
+					} else {
+						// Found a non-whitespace character. The original .trim() would result
+						// in a non-empty string. Return 0 as per original logic.
+						return 0;
 					}
 				}
+				// If the loop completes, all characters between start and index were checked whitespace.
```
pokoli, line 1729: Please remove non needed comments

pokoli (changes requested):

Ray0907:
> Thanks for the proposal.
> 
> Could you please provide some benchmark that proves your proposal makes the code faster?

I've tested both implementations across three file sizes.
```
File size: 0.02 MB
Origin (5 runs) | Avg Time: 0.58 ms | Avg Mem Δ: 0.18 MB | Rows: 100
Optimized (5 runs) | Avg Time: 0.51 ms | Avg Mem Δ: 0.17 MB | Rows: 100

File size: 16.52 MB
Origin (5 runs) | Avg Time: 142.94 ms | Avg Mem Δ: 73.69 MB | Rows: 100000
Optimized (5 runs) | Avg Time: 142.13 ms | Avg Mem Δ: 73.28 MB | Rows: 100000


File size: 333.23 MB
Origin (5 runs) | Avg Time: 3190.77 ms | Avg Mem Δ: 1423.03 MB | Rows: 2000000
Optimized (5 runs) | Avg Time: 3110.69 ms | Avg Mem Δ: 1423.30 MB | Rows: 2000000
```

Ray0907:
> Thanks for sharing your benchmarks but for me there is no performance gain as numbers are quite similar.

Yeah, the performance didn’t really get much better. Thanks ur feedback

PR #1095, opened 2025-05-19
Fix header name counting issue

aszx87410:
Using `__proto__` as a header name and repeating it can cause an infinite loop due to incorrect counting. This issue can be resolved by replacing {} with an object that has no prototype.


pokoli:
Thanks for submiting the fix. I released 5.5.3 version which contains this fix!

PR #1096, opened 2025-05-23
Mention BYTE_ORDER_MARK in documentation

forivall:
Adding this to the documentation so that it can be added to the definitely typed definition, as documentation indicates that it is a stable property.


dboskovic (approved):
💯

PR #1097, opened 2025-06-18
Update Node versions

gabegorelick:
Drop versions that are EOL and add newer versions.

Node 18 has been end of life since March 27, 2025

### DIFF
--- a/tests/test-cases.js
+++ b/tests/test-cases.js
@@ -13,6 +13,8 @@ var UNIT_SEP = String.fromCharCode(31);
 var FILES_ENABLED = false;
 try {
 	new File([""], ""); // eslint-disable-line no-new
+	// Required since Node20 as it ads a FileGlobal but not a FileReaderSync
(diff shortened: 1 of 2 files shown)


gabegorelick:
Not sure what these test failures are about. Will have to fix them.

gabegorelick:
Tests are fixed.

tests/test-cases.js
```diff
@@ -13,6 +13,7 @@ var UNIT_SEP = String.fromCharCode(31);
 var FILES_ENABLED = false;
 try {
 	new File([""], ""); // eslint-disable-line no-new
+	new FileReaderSync(); // eslint-disable-line no-new
```
gabegorelick, line 16: Node 20 added a File global, but not FileReaderSync or FileReader. We need those or else tests fail when we try to instantiate a FileReaderSync: https://github.com/gabegorelick/PapaParse/blob/be7e72cfd4baf3c25d8631b5823e1b4ae88364be/papaparse.js#L667-L681

Issue #751, opened 2019-12-10
ReadableStreamStreamer sometimes breaks UTF8-characters to ��


jehna:
## Issue

If I pass a CSV stream to `Papa.parse` that contains special characters, it sometimes breaks the special characters so they show up as e.g. ��.

## How to reproduce

See example at:
https://repl.it/repls/ArcticTreasuredLine

Press "Run" at the top of the page

### What should happen?

There should only be output of `ä` character

### What happened?

There's random occurrences of `��`

## Root cause

These two lines are responsible for this issue:
https://github.com/mholt/PapaParse/blob/master/papaparse.js#L863
https://github.com/mholt/PapaParse/blob/master/papaparse.js#L506

So when papaparse reads a chunk, it directly calls `.toString` of that chunk.

However a chunk consists of bytes, and some utf8-characters are two bytes long:
* `ä` consists of two bytes: `11000011` and `10100100`
* `a` (and other "regular" characters) is just one byte `01100001`

Now if the chunk splits right between a multi-byte character like `ä`, papaparse calls `toString` to both parts of the character distinctly, and produces two weird characters:

`11000011` (from end of first chunk) transforms to `�`
`10100100` (from start of second chunk) transforms to `�`

## How to fix this issue?

If received chunks of bytes, the concatenation should be done in bytes too, e.g. using `Buffer.concat`. Papaparse should not call `toString` before it has split the stream to lines, so the `_partialLine` remains as a buffer rather than a string type.

pokoli:
Hi @jehna, 

Thanks for reporting the issue. 

What you propose makes sense to me. It will be great if you can provide a PR that adds a test case for this issue and fixes it.

jehna:
## Workaround

I implemented a workaround for this issue: Use another library to pre-parse the lines as a stream.

I'm using [`delimiter-stream` NPM package](https://www.npmjs.com/package/delimiter-stream) that seems to have a correct implementation of line parsing as a byte stream:

https://github.com/peterhaldbaek/delimiter-stream/blob/043346af778986d63a7ba0f87b94c3df0bc425d4/delimiter-stream.js#L46

Using this library you can do a simple wrapper to wrap up your stream:

```js
const toLineDelimitedStream = input => {
  // Two-byte UTF characters (such as "ä") can break because the chunk might get
  // split at the middle of the character, and papaparse parses the byte stream
  // incorrectly. We can use `DelimiterStream` to fix this, as it parses the
[...]

PR #1099, opened 2025-07-09
Automatically resize chunks to prevent splitting UTF-8 characters

ManBearTM:
I took a stab at fixing https://github.com/mholt/PapaParse/issues/751 (and maybe the source of a few other issues). You could probably make a similar fix for UTF-16 files, but I opted to not look into that right now.

EDIT: I realize now that this fix maybe only applies to `FileStreamer`

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -739,6 +739,7 @@ License: MIT
 		ChunkStreamer.call(this, config);
 
 		var reader, slice;
+		var decoder = new TextDecoder(config.encoding || 'utf-8');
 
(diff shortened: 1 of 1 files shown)


dboskovic:
🤔 This logic makes sense but almost definitely adds a major performance hit. It seems the right solution would be to make sure the left over bytes properly re-assemble into valid utf8 bytecode. If you could add a regression test that validates this I'll include the fix in my #1100 refactor

ManBearTM:
Purely anecdotal, but I tested it with a 50 MB file and did not notice any performance hit. I can try to find the time to add tests, but I can't promise anything - life is a bit busy atm, sorry.

PR #1104, opened 2025-08-19
Update README.md: HTTP => HTTPS

Schweinepriester:
as one gets a warning now in chrome when opening a HTTP-only link

image

switch to HTTPS links.

as https://papaparse.com/ doesn't appear to have a valid cert

image

but https://www.papaparse.com/ does, immediately link there o7


pokoli:
Thanks for the update!

PR #1106, opened 2025-08-21
Add custom type cast function

rpwebdevelopment:
- Added custom type cast function option
- Updated docs to detail new option

New functionality provides a new option that allows users to define thir own dynamic type casting rules, this allows for more\less strict casting based upon the users specific requirements rather than the current more binary dynamicTyping. For example if the user only required boolean an null casting their config could be set up as below:

```
Papa.parse(file, {
    dynamicTyping: true,
    typeCastFunction: (val) => {
        if (val === 'true' || val === 'TRUE') {
            return true;
        } else if (val === 'false' || val === 'FALSE') {
            return false;
        }
        return (val === '' ? null : val);
    }
})
```


pokoli:
We have a `transform` [config option ](https://www.papaparse.com/docs#config) which can be applied to each value to cast to the required value. Reading your exapmle code it seems that just using your fuction as transform will work as what you need. 

So for me, what you are proposing is already implemented.

PR #1107, opened 2025-09-02
issue:998 Correctly identify the first chunk across pause/resume

daniele-pini:
### Brief description

This PR fixes unexpected header recognitions across calls to `parser.pause()` / `parser.resume()`.

Issue #998 was happening because the header recognition code was called multiple times for rows that were not the first, when the above mentioned functions were invoked.


papaparse.js
```diff
@@ -507,6 +507,8 @@ License: MIT
 
 		this.parseChunk = function(chunk, isFakeChunk)
 		{
+			var notFirstChunk = !this.isFirstChunk;
+
```
daniele-pini, line 511: Explaining the changes to facilitate the review. When parsing chunks, we want parsing to keep track of whether some previous chunk was already processed or not. This is basically to know whether header recognition should trigger.

We already have this information in the `ChunkStreamer` as the `isFirstChunk` property. We need to invert it before passing it down in order to make the default a _falsy_ value. This is used for example at this point: https://github.com/mholt/PapaParse/blob/b10b87ef8686c6f88299b50dd25e83606e9c36d4/papaparse.js#L1350-L1355

**NOTE:** I think the _first chunk_ may not necessarily _contain the header_, although it usually does. It depends on how big the header is and whether `skipFirstNLines` was used. In general, we could find the header on following chunks.

This is also a problem in current implementation, but fixing that would require further refactoring that I'm not comfortable carrying out myself. I would like to leave this problem to a future PR.

The correct way to refactor all this _really right_, I think, should be to move the [headerParsed](https://github.com/mholt/PapaParse/blob/b10b87ef8686c6f88299b50dd25e83606e9c36d4/papaparse.js#L1427) variable out of the `Parser` class somehow, because that one gets reinitialized all the time. A "parseContext" variable initialized in the highest level parse function could be used for that.

papaparse.js
```diff
@@ -1740,7 +1742,7 @@ License: MIT
 			/** Returns an object with the results, errors, and meta. */
 			function returnable(stopped)
 			{
-				if (config.header && !baseIndex && data.length && !headerParsed)
+				if (config.header && !notFirstChunk && data.length && !headerParsed)
```
daniele-pini, line 1745: The `notFirstChunk` variable gets passed down the parsing call chain up to this point. Previously, the `baseIndex` variable was used to determine whether to stop header recognition, while we now use the explicit `notFirstChunk` parameter.

The `baseIndex` (i.e. the cursor in the streamed file at the start of the chunk) was probably intended to be used for this role, except this doesn't work for the "fake chunks" generated when pausing and resuming. In particular, the first **real** chunk has a `baseIndex` of 0, and "fake chunks" inside it would also use that - which caused the header recognition to trigger multiple times.

PR #1119, opened 2026-03-13
Create CHANGELOG.md

vadimcoder:


pokoli:
Hi @vadimcoder, 

Did you write the changelog or you generate using any AI prompt? If the second will be great to share the prompt so we can continue updating with the same format in the future.

vadimcoder:
Hi @pokoli
I tried AI first, but it made some mistakes, I threw it away and just copy-pasted from here https://github.com/mholt/PapaParse/commits/master/ manually.

pokoli:
Great, so I will just take care of updating the file on release or even when accepting new PR. 

Thanks for the fast reply and the work on this!

Issue #1035, opened 2024-01-12
Non-quote quoteChar still quotes quotes


retorquere:
```
console.log(csv.unparse([['x"y']], { quoteChar: '@', delimiter: ' ' }))
```

should output

```
x"y
```

but instead outputs

```
@x"y@
```

PR #1124, opened 2026-06-18
fix: correctly escape and quote custom quoteChar in unparse

### LINKED ISSUE #1068: unparse: EscapeChar is wrongly set when quoteChar is changed
While looking into https://github.com/mholt/PapaParse/issues/1035 I noticed that the `escapeChar` is wrongly set when the `quoteChar` is changed. This is only the case for `Papap.unparse`.

Example: 

```js
Papa.unparse([['a', 'x+y']], {quoteChar: '+'})
```

which gives currently
```
a,+x""y+
```

As you can see, the `escapeChar` is still set to `""` and not `++`.

The issue was introduced in this commit: https://github.com/mholt/PapaParse/commit/a627547d87b6c155ccb46d61ee100c7a20581cf8#diff-fbb8d0b19dc4586fadd80f52a7b49415d2d41a5fdccaeaa728f7fca51cfbbb29R277

This can be easily be fixed when changing line https://github.com/mholt/PapaParse/blob/4af6882e0ea91b0baad30345f190b38d50e7535e/papaparse.js#L364C4-L364C29

```js
if (typeof _config.quoteChar === 'string')
  _quoteChar = _config.quoteChar;
```

into

```js
if (typeof _config.quoteChar === 'string') {
  _quoteChar = _config.quoteChar;
  _escapedQuote = _quoteChar + _quoteChar;
}
```

And here is a test-case for the `UNPARSE_TESTS`

```js
{
  description: "Other quote char but without setting escape char (should be 2x the quote char)",
  input: [['a', 'x+y']],
  config: { quoteChar: "+"},
  expected: 'a,+x++y+'
}
```

However, fixing this will result in
```
a,x++y
```

The correct result would be
```
a,+x++y+
```

The other part of the issue is that in line https://github.com/mholt/PapaParse/blob/4af6882e0ea91b0baad30345f190b38d50e7535e/papaparse.js#L472

```js
needsQuotes = needsQuotes
			|| _quotes === true
			|| (typeof _quotes === 'function' && _quotes(str, col))
			|| (Array.isArray(_quotes) && _quotes[col])
			|| hasAny(escapedQuoteStr, Papa.BAD_DELIMITERS)
			|| escapedQuoteStr.indexOf(_delimiter) > -1
			|| escapedQuoteStr.charAt(0) === ' '
			|| escapedQuoteStr.charAt(escapedQuoteStr.length - 1) === ' ';
```

Will always set `needsQuotes=true` if the field contains a `"` because `Papa.BAD_DELIMITERS` is defined as `['\r', '\n', '"', Papa.BYTE_ORDER_MARK]`.

There are different ways to fix this, here is one
- make a new constants, e.g. `Papa.NEED_QUOTES_CHARS=['\r', '\n']` and use it here instead of `Papa.BAD_DELIMITERS`
- add an additional check whether the field contains the actual quoteChar, e.g. `str.indexOf(_quoteChar) > -1`

These changed will also fix issue https://github.com/mholt/PapaParse/issues/1035

At this point, a PR would probably be easier... However, this can be implemented in different ways, so I'm hesitant.

chatman-media:
## Problem

When a custom `quoteChar` is passed to `Papa.unparse` (without an explicit `escapeChar`), two things go wrong:

1. The escaped-quote sequence (`_escapedQuote`) is left at its default `""` instead of being derived from the custom `quoteChar`, so an embedded quoteChar is escaped with the wrong character.
2. A field that contains the custom `quoteChar` is **not** detected as needing quotes, so it is emitted unquoted. Combined with (1), this produces output that cannot be parsed back — a round-trip loses data.

```js
Papa.unparse([['a', 'x+y']], { quoteChar: '+' })
// before: a,+x""y+      (wrong escape, and 2nd field not quoted -> reparses to x""y)
// after:  a,+x++y+        (round-trips back to x+y)
```

A clearer data-corruption case (quoteChar in the middle of a value):

```js
Papa.unparse([["ab'cd", 'next']], { quoteChar: "'", escapeChar: "'" })
// before: ab''cd,next     -> Papa.parse(...) === [['ab''cd','next']]   (corrupted)
// after:  'ab''cd',next    -> Papa.parse(...) === [['ab'cd','next']]    (correct)
```

This is the root cause reported in #1068 (and underlies the confusion in #1035).

## Fix

- `unpackConfig()` now sets `_escapedQuote = _quoteChar + _quoteChar` when `quoteChar` is configured. The existing explicit-`escapeChar` branch runs afterwards and still overrides it, so `escapeChar` behavior is unchanged.
- `safe()` now also marks a field as needing quotes when the value contains the `quoteChar` (generalizing the existing `BAD_DELIMITERS` check, which only covered the default `"`).

## Tests

- Added 3 unparse fixtures covering custom `quoteChar` with and without a custom `escapeChar`.
- Updated one existing fixture ("Escape formulae with single-quote quoteChar and escapeChar") whose expected output encoded the previous buggy, non-round-trippable behavior; the new expected output is RFC-consistent and round-trips correctly.
- Full suite green (`npm run test-node`, `npm run lint`).

Closes #1068

### DIFF
--- a/tests/test-cases.js
+++ b/tests/test-cases.js
@@ -2065,12 +2065,6 @@ var UNPARSE_TESTS = [
 		expected: 'a,+x++y+'
 	},
 	{
-		description: "Custom quoteChar quotes a field containing the quoteChar",
-		input: [['a@b', 'c']],
-		config: {header: false, quoteChar: '@'},
-		expected: '@a@@b@,c'
-	},
-	{
 		description: "Custom quoteChar with custom escapeChar quotes and escapes embedded quoteChar",
 		input: [['a', 'x+y']],
 		config: {header: false, quoteChar: '+', escapeChar: '\\'},
--- a/tests/test-cases.js
+++ b/tests/test-cases.js
@@ -2059,6 +2059,24 @@ var UNPARSE_TESTS = [
 		expected: 'foo,"""quoted"""'
 	},
 	{
+		description: "Custom quoteChar escapes and quotes embedded quoteChar (no escapeChar)",
+		input: [['a', 'x+y']],
+		config: {header: false, quoteChar: '+'},
+		expected: 'a,+x++y+'
+	},
+	{
+		description: "Custom quoteChar quotes a field containing the quoteChar",
+		input: [['a@b', 'c']],
+		config: {header: false, quoteChar: '@'},
+		expected: '@a@@b@,c'
+	},
+	{
+		description: "Custom quoteChar with custom escapeChar quotes and escapes embedded quoteChar",
+		input: [['a', 'x+y']],
+		config: {header: false, quoteChar: '+', escapeChar: '\\'},
+		expected: 'a,+x\\+y+'
+	},
+	{
 		description: "Escape formulae",
 		input: [{ "Col1": "=danger", "Col2": "@danger", "Col3": "safe" }, { "Col1": "safe=safe", "Col2": "+danger", "Col3": "-danger, danger" }, { "Col1": "'+safe", "Col2": "'@safe", "Col3": "safe, safe" }],
 		config: { escapeFormulae: true },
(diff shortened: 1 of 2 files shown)


tests/test-cases.js
```diff
@@ -2058,6 +2058,24 @@ var UNPARSE_TESTS = [
...
 		expected: 'foo,"""quoted"""'
 	},
+	{
+		description: "Custom quoteChar escapes and quotes embedded quoteChar (no escapeChar)",
+		input: [['a', 'x+y']],
+		config: {header: false, quoteChar: '+'},
+		expected: 'a,+x++y+'
+	},
+	{
+		description: "Custom quoteChar quotes a field containing the quoteChar",
```
pokoli: This tests seems exactly the same as above. So I do not think is needed.

pokoli (changes requested):

pokoli:
Thanks for your patch. Could you please remove the extra test so I can merge it? TIA

chatman-media:
Good catch — removed the redundant case (b036315). The `+`-quoteChar test above it already covers the same behavior (field containing the custom quoteChar, no escapeChar → field quoted + embedded quoteChar doubled), so the `@` variant was duplicative. Kept the `+` no-escapeChar case and the custom-escapeChar case. Tests green.

Issue #1111, opened 2025-10-24
False positive UndetectableDelimiter error when parsing single-column CSV


arberiii:
### Description
When parsing a valid single-column CSV string, PapaParse reports an unnecessary `UndetectableDelimiter` warning, even though parsing succeeds and the delimiter is not ambiguous.

### Example
```js
const text = `id\n1\n2\n3`

Papa.parse(text, {
  complete: (results) => {
    console.log(results)
  },
  error: (error) => {
    console.error('Failed to parse CSV:', error)
  }
})
```

Actual behavior

The parsed results are correct, but the output also includes an error:
```json
{
  "type": "Delimiter",
  "code": "UndetectableDelimiter",
  "message": "Unable to auto-detect delimiting character; defaulted to ','"
}
```

Expected behavior

For a single-column CSV with a valid header and data, PapaParse should not emit an UndetectableDelimiter error.
The defaulting behavior is fine internally, but this should not surface as an error in the results.

hiros0921:
I looked into this. The behaviour is deliberate rather than an oversight, so I did not open a PR — but there is a small change that would remove the false alarm, if you want it.

## Where it comes from

A delimiter only counts as detected when it produces more than about two fields per row on average:

https://github.com/mholt/PapaParse/blob/e6c2b7bc64a63addeac88c7b42c8d5d24f437beb/papaparse.js#L1294

A single-column input contains none of the candidate delimiters, so no candidate ever clears that bar, `bestDelim` stays undefined, and `UndetectableDelimiter` is reported even though the default delimiter parses the input correctly.

It is also pinned by the current test suite, which expects the error for these inputs:

https://github.com/mholt/PapaParse/blob/e6c2b7bc64a63addeac88c7b42c8d5d24f437beb/tests/test-cases.js#L1135

https://github.com/mholt/PapaParse/blob/e6c2b7bc64a63addeac88c7b42c8d5d24f437beb/tests/test-cases.js#L1155

https://github.com/mholt/PapaParse/blob/e6c2b7bc64a63addeac88c7b42c8d5d24f437beb/tests/test-cases.js#L1260

So changing it is a behaviour change, not a bug fix, and those three expectations would have to change with it.

## A possible change

Track whether any candidate delimiter split any row into more than one field. If none did, the input has no delimiter to detect and the default parses it correctly, so the error can be skipped:

```js
// in guessDelimiter()
var singleColumn = true;
...
    var fieldCount = preview.data[j].length;
[...]

PR #1125, opened 2026-06-22
fix: suppress false-positive UndetectableDelimiter error for single-column CSVs

JSap0914:
## Problem

When parsing a valid single-column CSV string without specifying `delimiter`, `Papa.parse()` emits a spurious `UndetectableDelimiter` error even though the data is parsed correctly.

```js
Papa.parse('id\n1\n2\n3')
// data: [['id'],['1'],['2'],['3']]  ← correct
// errors: [{ type: 'Delimiter', code: 'UndetectableDelimiter', ... }]  ← false positive
```

Reported in #1111.

## Root cause

`guessDelimiter()` only accepts a delimiter candidate when `avgFieldCount > 1.99`. A single-column CSV produces `avgFieldCount === 1` for every candidate delimiter, so no candidate is accepted and `_delimiterError` is set, triggering the error — even though the file is parsed correctly.

## Fix

Track a `singleColumnConsistent` flag inside `guessDelimiter()`: if any candidate delimiter produces perfectly consistent single-field rows (`delta === 0`, `avgFieldCount === 1`) across **more than one non-empty row**, the input is a valid single-column CSV, not a detection failure. In that case, return `successful: true` with the default comma delimiter.

Single-row inputs (`'Abc def'`) retain the existing error because a lone field is genuinely ambiguous.

## Tests

Two new test cases added to `tests/test-cases.js`:

- *Single-column CSV does not emit false-positive UndetectableDelimiter error*
- *Single-column CSV with header does not emit UndetectableDelimiter error*

All 249 passing tests continue to pass (`npm run test-node`); no existing tests were modified.

> AI-assisted contribution (GitHub Copilot / Claude)

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -1343,6 +1343,7 @@ License: MIT
 
 		function guessDelimiter(input, newline, skipEmptyLines, comments, delimitersToGuess) {
 			var bestDelim, bestDelta, fieldCountPrevRow, maxFieldCount;
+			var singleColumnConsistent = false;
 
 			delimitersToGuess = delimitersToGuess || [',', '\t', '|', ';', Papa.RECORD_SEP, Papa.UNIT_SEP];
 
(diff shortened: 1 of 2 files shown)


Audgui-Byte:
I tested this against current `master` (4843fc2) and this PR head (f23bc87). For `Papa.parse('a,b\nc;d')`, both return `[['a', 'b'], ['c;d']]`, but `master` reports `UndetectableDelimiter` while this PR returns `errors: []`. A candidate absent from both rows (such as a tab) gives consistent one-field rows, so `singleColumnConsistent` appears to suppress the warning even though comma and semicolon each split one row. Could the suppression require that *no* candidate splits any row, and could this mixed-delimiter case be added as a regression test? This matches the distinction discussed in #1111 between true single-column input and ambiguous delimiters. The branch also predates current `master`, so it will need updating before merge.

Issue #1103, opened 2025-08-12
parse stream will emit an error when newline is \r\n and the chunk data end with \r.


acrazing:
Code snippet:

```ts

import { Readable } from 'node:stream';
import Papa from 'papaparse';

const stream = new Readable({
  objectMode: true,
  read() {},
});

Papa.parse(stream, {
  delimiter: ',',
  newline: '\r\n',
  skipEmptyLines: true,
  header: true,
  complete(results, file) {
    console.log(results.errors);
  },
});

stream.push('name,value\r\n');
stream.push('foo,"FOO"\r');
stream.push('\n');
stream.push(null);

```

Got:

```ts
[
  {
    type: 'Quotes',
    code: 'InvalidQuotes',
    message: 'Trailing quote on quoted field is malformed',
    row: 0,
    index: 5
  }
]
```

Issue #636, opened 2019-02-19
Pause and resume (In a quick succession) gets you  lot of exceptions and an infinite loop


varunsharma27:
Doing pause and resume (In a quick succession) on demo gets you  lot of exceptions :- 

```
papaparse.js:733 Uncaught DOMException: Failed to execute 'readAsText' on 'FileReader': The object is already busy reading Blobs.
    at FileStreamer._readChunk (https://www.papaparse.com/resources/js/papaparse.js:733:21)
    at FileStreamer._nextChunk (https://www.papaparse.com/resources/js/papaparse.js:722:10)
    at FileStreamer.ChunkStreamer.parseChunk (https://www.papaparse.com/resources/js/papaparse.js:538:10)
    at FileStreamer._chunkLoaded (https://www.papaparse.com/resources/js/papaparse.js:743:9)
    at FileReader.<anonymous> (https://www.papaparse.com/resources/js/papaparse.js:1800:25)
FileStreamer._readChunk	@	papaparse.js:733
FileStreamer._nextChunk	@	papaparse.js:722
ChunkStreamer.parseChunk	@	papaparse.js:538
FileStreamer._chunkLoaded	@	papaparse.js:743
(anonymous)	@	papaparse.js:1800
load (async)		
FileStreamer.stream	@	papaparse.js:710
CsvToJson	@	papaparse.js:265
parseNextFile	@	papaparse.js:163
$.fn.parse	@	papaparse.js:113
(anonymous)	@	demo.js:103
dispatch	@	jquery.min.js:3
r.handle	@	jquery.min.js:3
```
And if you choose to handle them, like this :- 
```
function stepFn(results, parser)
{
	stepped++;
	if (results)
	{
		parser.pause();
		try{
		parser.resume();
		} catch (e) {
			console.log(e);
		}
		if (results.data){
			array2 = array2.concat(results.data); // Just to check the data
			rowCount += results.data.length;
		}
		if (results.errors)
		{
			errorCount += results.errors.length;
			firstError = firstError || results.errors[0];
		}
	}
}
```
Then you get into an infinite loop with the same data.

It appears that resume should be waiting for previous pause to halt the parsing before continuing again.
It appears this wasn't the case before since Issue #321 talks about doing this to produce another bug (Cursor related).
I am using :- 
`Papa.LocalChunkSize = 10485`
Otherwise you get :- 
```
RangeError: Maximum call stack size exceeded
[...]

PR #1127, opened 2026-06-26
fix: do not flag a malformed-quote error when a \r\n line ending is s…

Jian-Zhang08:
…plit across a chunk boundary (#1103)

When streaming, a chunk can end on the \r of a \r\n line ending, leaving a closing quote followed only by a lone \r. That \r matches neither the delimiter nor a full newline, so the parser reported a spurious "Trailing quote on quoted field is malformed" (InvalidQuotes) error even though the row is simply incomplete and is already deferred to the next chunk via _partialLine (no data is lost).

Guard the InvalidQuotes branch: when ignoreLastRow is set (more chunks are coming) and there is no further newline left in this chunk, defer the row instead of emitting the error. Genuine malformed quotes are unaffected (there a newline still follows, or it is the final chunk).

Adds a regression test that parses a quoted CSV with a \r\n line ending split across a chunk boundary and asserts no error is produced.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -1625,6 +1625,14 @@ License: MIT
 						}
 
 
+						// When streaming, a closing quote with no further newline left in this
+						// chunk means the row is still incomplete -- most likely a delimiter or
+						// newline split across the chunk boundary (e.g. a "\r\n" cut right after
+						// the "\r"). Defer the row to the next chunk instead of reporting a
+						// spurious malformed-quote error. (#1103)
+						if (ignoreLastRow && nextNewline === -1)
+							return finish();
+
 						// Checks for valid closing quotes are complete (escaped quotes or quote followed by EOF/delimiter/newline) -- assume these quotes are part of an invalid text string
 						errors.push({
 							type: 'Quotes',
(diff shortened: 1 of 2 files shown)


sridharkalaibala:
Independently checked head `24b58d7d550610bbd1af21961c19ed190bc5dfb7`:

- The new quoted-CRLF regression produces `InvalidQuotes` on the parent parser and parses cleanly with this change.
- 290 chunk-size cases passed across string chunking and Node Readable streams: quoted CRLF rows, headers/skipEmptyLines, embedded CRLF, escaped quotes, a custom quote character and multi-character delimiter. Malformed closing quotes still report `InvalidQuotes`, including an unterminated final row.
- The Node test files produced 247 passes and one 30-second timeout in the existing #636 pause/resume test. In isolation that test passed on both the parent (29.4s) and this head (29.2s), so I am not claiming a fully green single suite run. On Windows I first restored the CSV fixtures to their exact Git blob bytes to avoid checkout CRLF conversion changing the expected newline/cursor metadata.

This was a Node 24.19.0/Windows check; I did not run browser/worker tests. No additional correctness issue found in the cases above. Thanks for including the focused regression test.

Issue #1083, opened 2025-01-16
transformHeader Called Multiple Times in Papa Parse v5.5.1


sambit04126:
Bug Report: transformHeader Called Multiple Times in Papa Parse v5.5.1
1. Title
transformHeader Function Invoked for Data Rows in Papa Parse v5.5.1

2. Description
In Papa Parse version 5.5.1, the transformHeader function is being called for each data row's fields instead of exclusively for the header row. This results in unexpected transformations of data fields, leading to incorrect parsed output.

3. Expected Behavior
transformHeader should be invoked only once for each header field in the first row of the CSV.
Data rows should be processed without invoking transformHeader.
4. Actual Behavior
transformHeader is invoked for each field in data rows, causing data values to be transformed incorrectly.
This leads to incorrect and malformed rows in the transformation process.
5. Steps to Reproduce
Setup the Environment:

Install Papa Parse version 5.5.1.

`npm install papaparse@5.5.1`

Create a file named index.js and paste the following code:

```
const Papa = require('papaparse');

const csvContent = `Name,Age,City
John Doe,30,New York
Jane Smith,25,Los Angeles
,abc`;

function mapHeader(header) {
    const transformed = header.trim().toLowerCase();
    console.log(`Transforming header: "${header}" -> "${transformed}"`);
    return transformed;
}

function transformData(data) {
    console.log(`Transforming data: "${data}"`);
    return data;
}

let currentChunk = [];
const chunkSize = 2;

function processChunk(chunk) {
    console.log(chunk);
}

Papa.parse(csvContent, {
    header: true,                
    dynamicTyping: true,        
    skipEmptyLines: true,       
    transformHeader: mapHeader,  
    transform: transformData,   
    step: function (results, parser) {
        console.log('Parsed Row:', results.data);
        currentChunk.push(results.data);

        if (currentChunk.length === chunkSize) {
            processChunk(currentChunk);
            currentChunk = []; 
        }
    },
    complete: function () {
        console.log('Parsing Complete.');
        if (currentChunk.length > 0) {
            processChunk(currentChunk);
        } else {
            console.log('No remaining data to process.');
        }
    },
    error: function (error) {
        console.error('Parsing Error:', error.message);
    }
});
```
Run the script 

`node index.js`

Observe the Output:

With Papa Parse v5.5.1:

```
Transforming header: "Name" -> "name"
Transforming header: "Age" -> "age"
Transforming header: "City" -> "city"
Transforming header: "name" -> "name"
[...]

PR #1131, opened 2026-07-16
Fix: duplicate-header dedupe corrupts data rows under streaming backpressure

haggholm:
## Problem

When parsing a stream with `header: true` and a **slow / backpressured** consumer, PapaParse's duplicate-header dedupe (`email` → `email_1`, `""` → `_1`) is wrongly applied to **data rows**. Any data row with duplicate or blank cell values gets `_1`/`_2` suffixes silently written into its values.

### Reproduction (on current `master`)

```js
const Papa = require('papaparse');
let input = 'email,middle_name,address2,country\n';
for (let i = 0; i < 500; i++) input += `user${i}@example.com,,,US\n`;

const stream = Papa.parse(Papa.NODE_STREAM_INPUT, { header: true });
const rows = [];
const pump = () => {                       // slow reader => backpressure
  const row = stream.read();
  if (row === null) return void stream.once('readable', pump);
  rows.push(row);
  setImmediate(pump);
};
stream.once('readable', pump);
stream.on('end', () =>
  console.log(rows.filter(r => r.address2 === '_1').length)); // 484 — should be 0
stream.end(Buffer.from(input));
```

Every empty `address2` becomes `_1` (~484 of 500 rows here).

## Root cause

The dedupe runs in the low-level parser's `returnable()`, guarded by `!baseIndex` and a per-`Parser` `headerParsed` flag. But a fresh `Parser` (with `headerParsed` reset) is created for **every chunk and every pause/resume**. Under backpressure the streamer pauses mid-first-chunk and returns before `_baseIndex` advances past `0`; on resume, `!baseIndex && !headerParsed` is true again, so the dedupe re-fires — treating the next **data** row as a header and suffixing its duplicate/blank values.

## Fix

Gate the dedupe on whether the header has already been consumed. `_fields` is populated once and persists at the `ParserHandle` level across chunks and pause/resume (unlike the per-`Parser` `headerParsed`). `ParserHandle` sets `parserConfig.headerAlreadyParsed = _config.header && _fields.length > 0`, and `returnable()` also checks `!config.headerAlreadyParsed`. Legitimate renaming of a genuinely duplicated header row is unchanged.

## Test

Adds a Node streaming regression test in `tests/node-tests.js` that reproduces the corruption under backpressure (fails before this change, passes after) and asserts genuine duplicate-header renaming still works.

## Notes

Small and self-contained. Related to #1017 (duplicate headers across chunks) and #129 / #1083, but this targets specifically the pause/resume re-fire.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -1113,6 +1113,18 @@ License: MIT
 			if (_config.preview && _config.header)
 				parserConfig.preview++;	// to compensate for header row
 
+			// Once the header row has been consumed (_fields populated) it must
+			// never be re-derived. A fresh low-level Parser is created for every
+			// chunk AND every pause/resume, so the per-Parser `headerParsed` flag
+			// and the `!baseIndex` heuristic are not sufficient on their own: a
+			// pause during the first chunk (backpressure from a slow consumer)
+			// returns from parseChunk before `_baseIndex` advances past 0, so the
+			// resumed parse would otherwise treat the next DATA row as a header
+			// and apply duplicate-header suffixing (e.g. "" -> "_1") to its
+			// values. Gating on whether the header was already parsed prevents
+			// that corruption.
+			parserConfig.headerAlreadyParsed = _config.header && _fields.length > 0;
+
 			_input = input;
 			_parser = new Parser(parserConfig);
 			_results = _parser.parse(_input, baseIndex, ignoreLastRow);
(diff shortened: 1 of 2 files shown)


pokoli:
I'm closing this as I understand it had been fixed with #1130 
Reopen it if you don't agree.

Issue #1132, opened 2026-07-27
Four streaming defects: chunk-boundary UTF-8 corruption, cached line-ending guess, header dedupe on resume, and no backpressure under pause()


lenamonj:
Each item below has an offline reproduction. Note your CI is green at this commit, so none of this is visible from the suite as it stands.

- **Multi-byte UTF-8 split across a chunk boundary decodes to `U+FFFD`** (`papaparse.js:891`). Any non-ASCII file parsed with chunking can silently lose characters.
- **The line ending is guessed from a truncated first chunk and then cached for the whole file** (`papaparse.js:1091`). With `chunkSize: 1024` on a CRLF file this leaves a stray `\r` on every last field, with no error raised and a correct row count, so it looks like clean data.
- **Header dedupe re-runs on every resumed row and rewrites data in place** (`papaparse.js:1750`). This reproduces issues #985 and #998 verbatim.
- **`pause()` applies no backpressure** (`papaparse.js:846-856`). 100% of the stream buffers while paused; a 400 MB file under `--max-old-space-size=128` dies with `FATAL ERROR: JavaScript heap out of memory`, where the unpaused control completes at 78 MB RSS.

Two things I checked that are **not** broken, so nobody wastes time on them: quoted fields spanning chunk boundaries are exact at every chunk size across four corpora, and the pause/resume regression test that fails on Windows is a timer-resolution artifact (`setTimeout(3)` costs ~15.8 ms there, and `resume` schedules one timer per row), not a deadlock.

I have a patch taking the suite to 248 passing / 0 failing with lint clean. Several existing open PRs (#1099, #909, #1107, #989, #1017, #1131) cover adjacent ground and predate this. Happy to open a PR or leave it here.

pokoli:
Hi @lenamonj, 

I will be great if you can open A PR for each issue and include a tests case for each of them. 
Once done I will review and try to merge them. 

Thanks.

lenamonj:
Thanks - opened one per issue, each with a regression test that fails on master:

- #1133 multi-byte character split across a chunk boundary
- #1134 line ending guessed from a truncated chunk
- #1135 header de-duplication re-running on every resumed row
- #1136 `pause()` applying no backpressure to the source

Each branch is cut from master independently, so they can be reviewed and merged in any order, and all six pairwise merges are clean. Every PR body carries the red-green output showing the new test failing before the fix and passing after.

[...]

PR #1133, opened 2026-07-27
Fix multi-byte characters corrupted at a chunk boundary

lenamonj:
`ReadableStreamStreamer` decodes each incoming `Buffer` on its own:

```js
queue.push(typeof chunk === 'string' ? chunk : chunk.toString(this._config.encoding));
```

A UTF-8 sequence that straddles a chunk boundary is therefore split across two independent
`toString` calls and both halves come back as U+FFFD. The row count and the error list stay
clean, so nothing signals the corruption.

Minimal reproduction:

```js
var Papa = require('./papaparse.js');
var Readable = require('stream').Readable;

var cjk = String.fromCharCode(0x4e2d, 0x6587);    // two CJK ideographs, 3 UTF-8 bytes each
var bytes = Buffer.from('a,b\n1,' + cjk + '\n2,x\n', 'utf8');
var split = bytes.indexOf(0xe4) + 1;              // one byte into the first ideograph
var chunks = [bytes.subarray(0, split), bytes.subarray(split)], i = 0;
var source = new Readable({read: function() { this.push(i < chunks.length ? chunks[i++] : null); }});

Papa.parse(source, {
    step: function(r) { console.log(JSON.stringify(r.data)); }
});
// ["a","b"]
// ["1","��文"]     <- expected ["1","中文"]
// ["2","x"]
```

**The fix.** Hold one `string_decoder.StringDecoder` for the life of the stream so an
incomplete sequence is carried over to the next chunk, and flush it in `_streamEnd`. The
decode path sits behind the existing `typeof PAPA_BROWSER_CONTEXT === 'undefined'` guard the
file already uses in three places, so the browser bundle keeps calling `Buffer#toString`.
After `npm run build`, `grep -c string_decoder papaparse.min.js` returns 0, so the browser
bundle is unaffected in size and behaviour. `papaparse.min.js` is not included in this PR.

**The test.** `tests/node-tests.js`, "handles a multi-byte character split across a chunk
boundary". It feeds a two-buffer stream split one byte into a three-byte character and
asserts all three rows parse exactly.

Reported in #1132.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -774,6 +774,7 @@ License: MIT
 		var queue = [];
 		var parseOnData = true;
 		var streamHasEnded = false;
+		var decoder = null;
 
 		this.pause = function()
 		{
@@ -816,11 +817,29 @@ License: MIT
 			}
 		};
 
+		// A chunk boundary can fall inside a multi-byte character, so the bytes
+		// cannot be decoded one chunk at a time: an incomplete sequence has to be
+		// held back until the next chunk completes it, or both halves decode to
+		// U+FFFD and the field is silently corrupted.
+		this._decode = function(chunk)
+		{
+			if (typeof PAPA_BROWSER_CONTEXT === 'undefined')
+			{
+				if (decoder === null)
+				{
+					var StringDecoder = require('string_decoder').StringDecoder;
+					decoder = new StringDecoder(this._config.encoding || 'utf8');
+				}
+				return decoder.write(chunk);
+			}
+			return chunk.toString(this._config.encoding);
+		};
+
 		this._streamData = bindFunction(function(chunk)
 		{
 			try
 			{
-				queue.push(typeof chunk === 'string' ? chunk : chunk.toString(this._config.encoding));
+				queue.push(typeof chunk === 'string' ? chunk : this._decode(chunk));
 
 				if (parseOnData)
 				{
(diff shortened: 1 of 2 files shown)


lenamonj:
Friendly ping on this one. It still applies cleanly on master and CI is green; happy to rebase or trim it down if you'd prefer a smaller change.

sridharkalaibala:
I independently tested the decoder change at `586874acbf678915c111b611209f45a53d5d592a` with actual Node Readable inputs. All 26 checks passed:

- UTF-8 and UTF-16LE input containing Chinese text, emoji, an accented character, and a quoted embedded newline.
- Chunk sizes 1, 2, 3, 4, 7 and 16 bytes, each with normal step callbacks and with `pause()`/`setImmediate(resume)` on every row.
- An incomplete UTF-8 sequence at EOF, confirming the final replacement character is delivered by decoder flushing.
- Already-decoded string chunks with pause/resume.

For the byte-chunk cases I compared complete rows and parse errors against parsing the same bytes decoded as one string. Running the same probe against this commit's parent (`c26a506`) fails on the first one-byte UTF-8 case: `中文 😀 café` becomes replacement characters, while the error list remains empty.

I supplied an explicit newline to isolate this from the separate line-ending detection issue. This is additional validation of the Readable decoder path, not a full-suite, browser, backpressure, or `NODE_STREAM_INPUT` sign-off.

PR #1136, opened 2026-07-27
Fix pause() applying no backpressure to the source

lenamonj:
`ReadableStreamStreamer.pause` and `.resume` start with

```js
ChunkStreamer.prototype.pause.apply(this, arguments);
```

but `ChunkStreamer` assigns every method to `this` and leaves its prototype empty, so both
throw `TypeError: Cannot read properties of undefined (reading 'apply')` on their first line.
Nothing calls them, which is why this has never surfaced. The consequence is that `pause()`
never throttles the source: `ParserHandle.pause` stops the parser but the stream keeps
emitting, and `_streamData` keeps decoding and queueing, so the whole rest of the input piles
up in memory for as long as the caller stays paused. That is the case `pause()` exists to
avoid.

Measured on an 8 MB source with the parser paused after the first row, sampled 250 ms later:

```
pushed while paused: 8372000 bytes (7.98 MB)      master
pushed while paused:   13000 bytes (0.01 MB)      this branch, one read already in flight
```

**The fix.** Give `ChunkStreamer.prototype` real no-op `pause`/`resume` hooks that a streamer
with a throttleable source overrides, and have `ParserHandle.pause`/`resume` call them.

This PR also corrects `StringStreamer.prototype = Object.create(StringStreamer.prototype)` to
name `ChunkStreamer.prototype`. That self-assignment is inert while `ChunkStreamer.prototype`
is empty, but it is load-bearing the moment anything is put on it: without that one-word
change every existing string pause/resume test fails with
`TypeError: self.streamer.pause is not a function`. It is included here rather than split out
because this change is what makes it matter.

The change is deliberately narrow. It does not touch `parseChunk`, the `setTimeout(resume, 3)`
polling, or `meta.cursor`; those are separate problems.

**The test.** `tests/node-tests.js`, "pause() applies backpressure to the source stream". A
`Readable` counts the bytes it has pushed; the parser pauses at the first row and 250 ms later
the test asserts no more than two 13 KB blocks were pushed in the meantime. Bounded and fast:
the source is capped at 2 MB and destroyed at the end, so nothing large is allocated.

Reported in #1132.

**Red-green evidence:**

```
$ git checkout master -- papaparse.js
$ npx mocha tests/node-tests.js --grep "backpressure to the source"

  1) PapaParse
       pause() applies backpressure to the source stream:

      Uncaught AssertionError [ERR_ASSERTION]: source pushed 2080000 bytes while the parser was paused
      + expected - actual
      -false
      +true
```

```
$ git checkout fix/pause-backpressure -- papaparse.js
$ npx mocha tests/node-tests.js --grep "backpressure to the source"

  PapaParse
    [pass] pause() applies backpressure to the source stream (266ms)

  1 passing (268ms)
```

Confirming the `StringStreamer.prototype` line is load-bearing rather than cosmetic, with the
rest of the branch applied and only that line reverted:

```
$ npx mocha tests/test-cases.js --grep "Step functions can pause parsing"

  1) Custom Tests
       Step functions can pause parsing:
     TypeError: self.streamer.pause is not a function
      at ParserHandle.pause (papaparse.js:1141:18)
```

Suite and lint on the branch:

```
$ npm run lint      # exit 0, no output
$ npm run test-node
  248 passing (32s)
  23 pending
  1 failing         # the pre-existing bug #636 timeout, unchanged
```


pokoli:
@lenamonj tests are failing, could you have a look? Thanks

lenamonj:
Thanks for the ping. The failure is in my test, not in the fix.

A paused Readable still fills its own buffer up to the high-water mark, which is correct backpressure. The test hardcoded an allowance of two 13 KB blocks, which quietly assumed Node's old 16 KB default. That default is now 64 KB, so the source legitimately buffers 65000 bytes and the assertion trips. It passes under the old default and fails under the new one, which is why it looked environment specific.

I have pinned highWaterMark on the test's source and derived the allowance from it, so the bound no longer depends on the runtime default. The assertion is not loosened: with the fix removed the source pushes the full 2080000 bytes against an allowance of 29384.

Verified on this branch merged with master: 256 passing, 0 failing, lint clean.

lenamonj:
Correction to my previous comment. I attributed the differing high-water mark to a Node version change, and that is not what the measurements show. It differs by platform, not version:

```
v22.22.0  win32  16384
v22.22.0  linux  65536
v24.17.0  linux  65536
v26.0.0   linux  65536
```

So the old allowance encoded the 16 KiB default I develop against, while the CI runners use 64 KiB. The patch and the conclusion are unchanged: pinning highWaterMark in the test removes the dependency either way.

lenamonj:
Friendly ping. The test now accounts for the per-platform high-water mark and CI is green. Anything else you'd like adjusted?

PR #1137, opened 2026-07-30
Remove jQuery plugin integration

### LINKED ISSUE #841: remove jquery stuff
I think this tool should be as vanilla as possible and only do one thing

that means removing this jQuery bit
https://github.com/mholt/PapaParse/blob/5747da6c9927642743d6d44d49689be2d2fa0b10/papaparse.js#L87-L179

most ppl have moved away from the good old days of jquery now where MVC have taken over such as vue, angular, react, sapper etc

Kocayilmaz:
Removes the optional jQuery plugin (`$.fn.parse`) from papaparse.js, as requested in #841.

The library advertises zero dependencies, and this was the one block of code coupled to a specific DOM/library environment. It has no test coverage of its own, and the same behavior (parse files from a file input) is just as easy to get by wiring the input's change event to `Papa.parse()` directly.

Also dropped the `jquery-plugin` npm keyword and the README bullet mentioning it.

Ran the full node test suite before and after: 244 passing / 4 failing in both cases, so this doesn't change test results. The 4 failures are pre-existing and about CRLF vs LF linebreak detection on the sample CSV fixtures, unrelated to this change.

Fixes #841


pokoli:
Thanks for the MR, I did a first quick look and it seems good to me. 
I'm not sure if anyone is still using the jquery plugin but as this may be a broken change I plan to relase this as a new feature version (but relasing bug fixes first).

PR #1138, opened 2026-08-04
Add downloadTimeout option to abort slow remote requests

### LINKED ISSUE #786: timeout for download

Kocayilmaz:
Fixes #786.

Adds a `downloadTimeout` config option (milliseconds) for `download: true` requests. When set, it configures `XMLHttpRequest.timeout` on the `NetworkStreamer` request and surfaces a clear timeout error through the `error` callback if the remote file doesn't finish downloading in time. The value is validated with `parseInt`, following the same validation pattern used elsewhere in `CsvToJson` (throws a descriptive `Error` if not parsable).

There was a previous attempt at this in #789, which got a positive review but was abandoned by the author. This re-implements it against current master (the codebase has changed since 2020) and adds test coverage:

- Config validation: throws for non-numeric `downloadTimeout`, coerces numeric strings.
- A request that legitimately exceeds `downloadTimeout` and produces a timeout error.
- A request that finishes well within a generous `downloadTimeout`, to make sure normal downloads aren't affected.

The timeout test needed a way to reliably produce a slow response without depending on real network flakiness, so I added a small artificial delay in the local test server (`tests/test.js`) specifically for `tests/slow-sample.csv`.

Also documented the new option in `docs/docs.html`.


pokoli:
Thanks @Kocayilmaz, everything looks good to me. 
I will review some fixes MR (so I can publish a patch version with them) and latter include a minor version with this new feature!

PR #1142, opened 2026-08-16
Strip the UTF-8 BOM from streamed input, not just string input

dngr2:
The UTF-8 byte order mark is stripped from string input only. Every other input reaches the parser through `ChunkStreamer.parseChunk`, which never sees `stripBom`, so the mark stays inside the first field of the first row:

```js
const readStream = fs.createReadStream('utf-8-bom-sample.csv', 'utf8');
readStream.pipe(Papa.parse(Papa.NODE_STREAM_INPUT))
  .on('data', row => console.log(row));

// ['﻿A', 'B', 'C']   expected ['A', 'B', 'C']
```

Same for a `File`, a `download`, and a Node `Readable`. Nothing renders the character, so the first field simply never compares equal to what it looks like — `row[0] === 'A'` is false, and a `dynamicTyping` number in that position stays a string.

`CsvToJson` strips it at

https://github.com/mholt/PapaParse/blob/master/papaparse.js#L148

but that branch is guarded by `typeof _input === 'string'`.

### Why it has not shown up

`header: true` hides it. 08265f1 added a separate `stripBom` for column keys, so the header row is cleaned wherever it comes from, and only `header: false` is left exposed.

The existing coverage misses it for the same reason the bug is easy to miss. Both BOM cases in `test-cases.js` pass a string literal, and `tests/utf-8-bom-sample.csv` — a fixture that exists precisely for this — is only ever read into a string before being handed to `Papa.parse`. The streaming path has no BOM coverage at all.

### The change

Strip it once at the front of the stream, in the first-chunk block of `parseChunk`, which is where every input path converges. Doing it per streamer would mean four copies.

A mark anywhere other than the very start is data and is left alone.

### Tests

Three added to `tests/node-tests.js`, using the existing fixture through the pipe idiom already used there.

Worth being explicit about which one does the work: only **"strips the utf-8 BOM from a piped stream"** fails without the change. The `header: true` variant passes either way — it is there to document that the header path was already handled by 08265f1 — and "keeps a BOM character that is not at the start of the input" is a guard against over-stripping. `npm run lint && npm run test-node` gives 253 passing.

`papaparse.min.js` is left untouched, since it does not appear in contributor commits.

### DIFF
--- a/papaparse.js
+++ b/papaparse.js
@@ -422,6 +422,13 @@ License: MIT
 		this.parseChunk = function(chunk, isFakeChunk)
 		{
 			// First chunk pre-processing
+			// A byte order mark is an encoding artifact rather than data. String input
+			// is stripped in CsvToJson, but every other input reaches the parser through
+			// here, so strip it at the front of the stream too. Without this the mark
+			// stays inside the first field of the first row, where it is invisible.
+			if (this.isFirstChunk)
+				chunk = stripBom(chunk);
+
 			const skipFirstNLines = parseInt(this._config.skipFirstNLines) || 0;
 			if (this.isFirstChunk && skipFirstNLines > 0) {
 				let _newline = this._config.newline;
--- a/papaparse.js
+++ b/papaparse.js
@@ -145,7 +145,6 @@ License: MIT
 		}
 		else if (typeof _input === 'string')
 		{
-			_input = stripBom(_input);
 			if (_config.download)
 				streamer = new NetworkStreamer(_config);
 			else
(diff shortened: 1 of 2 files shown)


pokoli:
Hi @dngr2, 

First of all, thanks for your proposal!

It is possible to find a single path to strip the BOM? I think this will make much more sense than just adding the StripBOM on each code path.

dngr2:
Good call — consolidated. Every input type (string, stream, download, File) already flows through `parseChunk`, so I dropped the separate `stripBom` in `CsvToJson` and kept the single strip at the front of the first chunk. The existing string-input BOM test still passes via that path, alongside the new stream cases.

PR #1153, opened 2026-09-12
docs: remove duplicate keys from header result example

### LINKED ISSUE #1062: Documentation: Example objects contain duplicate keys
The "Data" section of https://www.papaparse.com/docs#results contains something that I think was not intentional on the part of the author.

When showing how the `header: false` behavior differs from the `header: true` behavior, it shows an example data structure in each case.

In the `header: true` example (linked below), the objects contains properties that do not have counterparts in the `header: false` example, and those properties happen to have the same keys as other properties in the same object. I assume this was the result of a copy/paste error at some point.

https://github.com/mholt/PapaParse/blob/a116779198bc0b067a13aa859c32731bdee7a4a1/docs/docs.html#L768-L780

I propose this as a fix:

```diff
 // Example (header: true) 
 [ 
 	{ 
 		"Column 1": "foo", 
 		"Column 2": "bar", 
- 		"Column 1": "foo1", 
 	}, 
 	{ 
 		"Column 1": "abc", 
 		"Column 2": "def", 
- 		"Column 1": "abc1", 
 	} 
 ]</code></pre> 
```

sridharkalaibala:
## Summary

Remove the duplicate `Column 1` properties from the `header: true` results example. They overwrite `foo` and `abc` with values absent from the adjacent `header: false` example, so the two examples currently describe different data.

Fixes #1062.

## Validation

- Extracted and executed both documented arrays and compared them with actual `Papa.parse` output for the same CSV. The original `header: true` example fails; both corrected examples pass.
- Node test suite: 255 passed, 25 pending. CSV fixtures were read with their committed line endings during this Windows run and restored afterward.
- `git diff --check` passed. Documentation-only change; parser source is unchanged.


lbesecker195 (approved):
Duplicate keys in a JSON object literal are a documentation error, and the extra values did not exist in the adjacent `header: false` example. LGTM.

sridharkalaibala:
Thanks for reviewing, @lbesecker195. Both corrected examples now match the parser output for the same input; the change stays limited to the documentation.

pokoli (approved):

PR #1154, opened 2026-09-12
docs: clarify default local file encoding

### LINKED ISSUE #1063: Please indicate in documentation what default value is for config.encoding
It would be useful to know what happens with ` { encoding: '' } `. Is encoding detected automatically, or is the default utf-8 or something similar. 

Many thanks!

sridharkalaibala:
## Summary

Clarifies that leaving `encoding` unset or using an empty string delegates local-file decoding to FileReader, with UTF-8 fallback and Unicode BOM handling. Explains that Papa Parse does not guess arbitrary encodings, gives Windows-1252 as an explicit example, and notes that already-decoded string input is unaffected.

Fixes #1063.

## Validation

- Traced FileStreamer to its `FileReader.readAsText` call and checked the File API decoding rules linked in the guide.
- Ran an isolated Chromium test against current `papaparse.js` with seven cases: omitted/empty encoding for UTF-8, explicit Windows-1252, undecorated Windows-1252 with fallback, UTF-16LE BOM, Blob MIME charset behavior, and string input.
- All cases passed. The MIME case confirmed sliced file blobs lose the charset, so the guide does not promise MIME charset detection.
- `git diff --check` passed. Documentation-only change; full parser suites were not rerun for this paragraph.


pokoli:
I do not think we need to document so much details.

sridharkalaibala:
Understood, thanks for reviewing it. I’ll keep future API documentation proposals concise and focused on the behavior users need to choose the option.

